PR #353, opened 2026-08-16
fix(starter): show the resolved compiler mode

Kinfe123:
## What changed

- Expose the resolved compiler flag through Farm public env.
- Render the actual enabled/disabled value in the starter UI.
- Give the disabled state a distinct neutral indicator.
- Make the production experiment aware of both compiler-on and compiler-off builds.
- Document that the visible status follows the build.

## Why

`FARM_REACT_COMPILER=false` disabled compilation but the page still claimed `experimental.compiler: true`. The UI now reads the same resolved value as the renderer configuration through Farm public env, which is available consistently during SSR and hydration.

## Validation

- `pnpm --filter @farm.js/create-app test`
- `pnpm exec oxfmt --check` on the changed template and test files
- The standalone mirror validates full compiler-on and compiler-off production experiments.

### DIFF
--- a/packages/create-farm-app/templates/react-compiler/README.md
+++ b/packages/create-farm-app/templates/react-compiler/README.md
@@ -37,6 +37,8 @@ FARM_REACT_COMPILER=true pnpm build
 FARM_REACT_COMPILER=false pnpm build
 ```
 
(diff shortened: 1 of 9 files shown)


Kinfe123:

PR #354, opened 2026-08-16
feat(react): verify compiler coverage reports

Kinfe123:
## What changed

- enable deterministic compiler coverage reports in the maintained React compiler example
- make the production experiment validate the generated report
- include the report path and summary in experiment output

## Why

The browser experiment proves visible behavior, while the report proves that intended components actually took the AOT path. Together they prevent silent compiler coverage regressions.

## Published starter boundary

The npm starter still uses @farm.js/react 0.1.0-beta.0, which predates report support. This PR intentionally leaves the generated and standalone starters unchanged. They can adopt the same report check after a renderer beta containing report support is published.

## Validation

- pnpm --filter @farm.js/react test (42 passed)
- pnpm --filter @farm.js/react build
- pnpm format:check
- FARM_EXPERIMENT_SCREENSHOT=/tmp/farm-react-compiler-report-merge.png pnpm --dir examples/react-compiler experiment
- report summary: 9 considered, 4 compiled, 5 fallback

### DIFF
--- a/examples/react-compiler/farm.config.ts
+++ b/examples/react-compiler/farm.config.ts
@@ -6,7 +6,7 @@ const compilerEnabled = process.env.FARM_REACT_COMPILER !== "false";
 export default defineConfig({
   renderer: react({
     experimental: {
-      compiler: compilerEnabled,
+      compiler: compilerEnabled ? { report: true } : false,
     },
   }),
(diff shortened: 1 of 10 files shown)


Kinfe123:

PR #355, opened 2026-08-16
feat(react): compile safe derived local values

Kinfe123:
## What changed

- accepts compiler-safe derived const values after top-level useState declarations
- expands chained derived values into DOM bindings and event handlers in source order
- falls back with focused diagnostics for forward references and expressions whose evaluation or identity is unsafe to lower
- exercises derived text and class bindings in the maintained production browser experiment
- documents the supported expression boundary

## Why

Common components name state-derived values before returning JSX. Keeping these components on the compiled path removes an artificial source-shape restriction without weakening the existing React fallback boundary.

## Impact and safety

The compiler only accepts statically analyzable expressions. Calls, assignments, functions, JSX, constructors, object and array literals, and similar unsupported expressions remain React-owned. Existing unsupported components continue to fall back rather than being partially compiled.

## Validation

- pnpm --filter @farm.js/react test (46 passed)
- pnpm --filter @farm.js/react type-check
- pnpm --filter @farm.js/react build
- production browser experiment passed with zero compiled update executions and no browser errors
- oxfmt check and git diff --check

---
## Summary by cubic
Compiles pure, state‑derived local const values declared after top‑level `useState`, removing an artificial fallback. Previously, any derived locals forced React fallback; now safe derived values inline into DOM bindings and event handlers in source order without creating a runtime scope.

- Compiler: validates and expands derived locals; enforces state‑before‑derived ordering; requires single‑identifier `const` and unique names; allows literals, props, state, operators, member access, conditionals, templates, and earlier derived values; rejects calls, assignments, array/object literals, functions, JSX, constructors, tagged templates, `await`, `this`/`delete`/update/`yield`; forbids forward references; `useState` initializers cannot reference any other local binding; removes derived identifiers from emitted code while preserving dependency tracking.
- Docs/examples/experiment: updates docs and `packages/farm-react/README.md` to include the derived‑locals boundary and examples (`statusClass`/`label`, `statusText`); `examples/react-compiler` now derives these values and the verifier asserts “DOM bindings patched.”
- Tests: adds success coverage for compiled derived locals and failure cases with focused diagnostics (impure calls, forward references, state after derived). No migration required; keep all `useState` declarations before derived locals.

<sup>Written for commit f49a9d682901918a310680c5522f07b86ede35cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/src/app/docs/renderers/react/page.md
+++ b/docs/src/app/docs/renderers/react/page.md
@@ -327,7 +327,7 @@ satisfy all of these rules:
 | ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Component discovery | A top-level, capitalized function declaration, function expression, or arrow component in application `.tsx` or `.jsx`.                             |
 | Function shape      | Synchronous, non-generator, non-generic block body with zero parameters or one identifier such as `props`.                                          |
-| Body                | One or more top-level `useState` declarations followed by one unconditional JSX return.                                                             |
+| Body                | Top-level `useState` declarations, optional compiler-safe derived `const` values in source order, then one unconditional JSX return.                |
 | State               | `const [value, setValue] = useState(initial)`, including lazy initializers, multiple cells, and queued functional updates.                          |
 | Root                | Exactly one lowercase host JSX element such as `button`, `section`, `input`, or `div`.                                                              |
 | Tree                | A static tree containing host elements only. Nested host elements are supported when their placement cannot change.                                 |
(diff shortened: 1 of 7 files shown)


Kinfe123:

PR #436, opened 2026-08-22
fix(core): read zod 4 schema internals in the openapi generator

### LINKED ISSUE #430: OpenAPI generator reads Zod 3 internals: numbers and enums become strings, arrays throw
## Summary

`OpenAPIGenerator` reads Zod schemas through private `_def` fields that only exist in
Zod 3. `packages/farm/package.json` depends on `zod: "^4.1.12"` (resolved 4.1.12), so
every read misses.

The result: numbers, booleans and enums are all documented as `string`, enum values are
dropped, and any array schema throws.

Everywhere else the framework stays version-agnostic by using the public API. For example
`validateInput` in `src/api/runtime.ts` calls `schema.parse()` and reads both
`validationError.errors` (v3) and `.issues` (v4). `src/openapi/generator.ts` is the only
file in `packages/farm/src` that touches `_def.typeName`.

## Impact

The OpenAPI document is wrong for almost every schema, and it fails quietly.

`z.array(...)` throws `TypeError: Cannot use 'in' operator to search for 'minLength' in array`.
Both callers catch it and carry on:

- `getParameters` (`generator.ts:469`) catches and returns what it collected so far, so the
  array parameter **and every parameter after it** disappear from the spec.
- `getRequestBody` (`generator.ts:401`) catches and falls through to the generic fallback at
  `generator.ts:407`, so the whole body becomes `{ type: "object", description: "Request body" }`
  and **all** properties are lost because one field was an array.

The example in the OpenAPI docs page is affected too. `docs/src/app/docs/openapi/page.md`
shows:

```ts
query: z.object({
  limit: z.coerce.number().int().positive().default(20),
}),
```

and says "The route appears in the generated reference with a typed `limit` query parameter".
It is emitted as `{"type":"string"}`.

## Reproduction

`processZodType` is private, so this calls it directly. Both the test and the generator get
the same `zod` instance through Vitest, so the `instanceof` checks behave normally.

```ts
import { describe, it } from "vitest";
import { z } from "zod";
import { OpenAPIGenerator } from "../openapi/generator";

function processed(schema: any) {
  const generator = new OpenAPIGenerator("/tmp", { title: "t" });
  try {
    return JSON.stringify((generator as any).processZodType(schema));
  } catch (error) {
    return `THREW: ${(error as Error).message}`;
  }
}

describe("openapi zod reading", () => {
  it("prints what the generator produces", () => {
    console.log("number      ->", processed(z.number()));
    console.log("boolean     ->", processed(z.boolean()));
    console.log("enum        ->", processed(z.enum(["asc", "desc"])));
    console.log("number.min  ->", processed(z.number().min(1).max(9)));
    console.log("docs limit  ->", processed(z.coerce.number().int().positive().default(20)));
    console.log("array       ->", processed(z.array(z.string())));
    console.log("object+array->", processed(z.object({ name: z.string(), tags: z.array(z.string()) })));
  });
});
```

Output on `zod@4.1.12`:

```
number      -> {"type":"string"}
boolean     -> {"type":"string"}
enum        -> {"type":"string"}
number.min  -> {"type":"string"}
docs limit  -> {"type":"string"}
array       -> THREW: Cannot use 'in' operator to search for 'minLength' in array
object+array-> THREW: Cannot use 'in' operator to search for 'minLength' in array
```

Expected: `{"type":"number"}`, `{"type":"boolean"}`,
`{"type":"string","enum":["asc","desc"]}`, `{"type":"number","minimum":1,"maximum":9}`, and
an array schema instead of a throw.

## Cause

Three reads in `packages/farm/src/openapi/generator.ts`.

**1. The type tag (`generator.ts:48`)**

```ts
const typeName = (zodType as any)._def?.typeName;
if (typeName) {
  if (typeName === "ZodString") return "string";
  ...
}
return "string";
```

Zod 4 has no `_def.typeName`. It uses `_def.type` with lowercase names, so `typeName` is
always `undefined` and every primitive falls through to `return "string"` at
`generator.ts:56`.

**2. The array element (`generator.ts:98`)**

```ts
const itemType = (zodType as any)._def.type;
return { type: "array", items: this.processZodType(itemType), ... };
```

In Zod 4, `_def.type` on an array is the string `"array"`; the element schema is at
`_def.element`. So `processZodType` recurses on a plain string, reaches the constraint block
at `generator.ts:122`, and `"minLength" in "array"` throws because `in` does not accept a
primitive.

**3. The enum values (`generator.ts:110`)**

```ts
enum: (zodType as any)._def.values,
```

`_def.values` is `undefined` in Zod 4, so `enum` is dropped from the JSON and the allowed
values are lost.

For reference, the shapes in both majors:

| read | zod 3 | zod 4 |
| --- | --- | --- |
| type tag | `_def.typeName` = `"ZodNumber"` | `_def.type` = `"number"` |
| array element | `_def.type` (schema) | `_def.element` (schema) |
| enum values | `.options` and `_def.values` | `.options` only |
| string min/max | `.minLength` / `.maxLength` | same |
| number min/max | `.minValue` / `.maxValue` | same |

The `instanceof z.ZodObject` / `ZodArray` / `ZodEnum` checks are fine in both, which is why
the array branch is entered at all.

Separately, the numeric constraints at `generator.ts:128-133` read `.min` / `.max`, which are
methods in both majors, so `baseSchema.minimum` is assigned a function and `JSON.stringify`
drops it. Numeric constraints have therefore never reached the document. `.minValue` /
`.maxValue` hold the values, in both majors.

## Suggested fix

Read the shapes both majors expose, so the generator does not depend on one:

- type tag: `_def.type` (v4) or `_def.typeName` (v3)
- array element: `_def.element ?? _def.type`
- enum values: `.options ?? _def.values`
- numeric constraints: `.minValue` / `.maxValue`
- guard the constraint block so a non-object value can never reach `in` again

`src/__tests__/openapi-generator.test.ts` currently has one test, for a route with no
schemas at all, so none of this is covered. I would add cases for primitives, enums, arrays,
nested objects and constraints.

## Environment

- `@farm.js/core` `0.1.0-beta.55`
- `zod` 4.1.12 (declared `^4.1.12`)
- Node.js 20.19.0, Windows 11

---

@Kinfe123 I'd like to work on this one. Could you assign it to me?

Kinfe123:
## Summary

Fixes #430.

`OpenAPIGenerator` read Zod 3 private fields while the package ships `zod ^4.1.12`, so the generated document was wrong for nearly every schema:

- `_def.typeName` no longer exists → numbers, booleans, and enums were all documented as `string`
- `_def.type` on a Zod 4 array is the string `"array"`, not the element schema → every array schema threw, and the two callers swallow the error, silently dropping parameters or the whole request body
- `_def.values` is gone → enum values were dropped
- numeric bounds were read from `.min`/`.max`, which are builder methods, so numeric constraints never reached the document in either major

## Fix

Read the shapes both majors expose:

- type tag: `_def.type` (v4) with `_def.typeName` (v3) fallback
- array element: `_def.element ?? _def.type`
- enum values: public `.options`
- numeric bounds: `.minValue`/`.maxValue`, skipping the implicit ±`MAX_SAFE_INTEGER` bounds Zod 4's `.int()` adds
- new `ZodDefault` branch so e.g. the docs example `z.coerce.number().int().positive().default(20)` emits `{type: "number", default: 20}`

## Tests

Added 9 cases to `openapi-generator.test.ts` covering primitives, enums, string/number constraints, arrays, nested objects, optionals, defaults, and descriptions. 8 of them fail before the fix (matching the issue repro exactly); all pass after.

---
## Summary by cubic
Fixes incorrect OpenAPI output by reading `zod` v4 schema internals instead of v3 private fields. Previously numbers/booleans/enums were emitted as string, arrays threw and dropped parameters or request bodies, and enum values and numeric constraints were omitted; now the generator emits correct types, array items, enum values, numeric bounds, and defaults.

**Review notes**
- Type detection: use `_def.type` (v4) with `_def.typeName` (v3) fallback.
- Arrays: read element via `_def.element ?? _def.type`.
- Enums: use public `.options` to include values.
- Numbers: read `.minValue`/`.maxValue`; ignore implicit ±MAX_SAFE_INTEGER from `.int()`.
- Defaults: unwrap `ZodDefault` to retain inner schema and default value; preserve descriptions and mark optionals as nullable.
- Tests: add 9 cases covering primitives, enums, constraints, arrays, nested objects, optionals, defaults, and descriptions.

<sup>Written for commit 060d1f3ced69af7e739cad264a9e3058eb51d2ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/436?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/openapi-generator.test.ts
+++ b/packages/farm/src/__tests__/openapi-generator.test.ts
@@ -2,6 +2,7 @@ import { mkdtempSync, realpathSync, rmSync, writeFileSync } from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
+import { z } from "zod";
 import { OpenAPIGenerator } from "../openapi/generator";
 
 const tempDirs: string[] = [];
@@ -43,6 +44,80 @@ describe("OpenAPIGenerator", () => {
           },
         },
       },
+    });
+  });
+});
+
+describe("OpenAPIGenerator zod schema conversion", () => {
+  function processed(schema: z.ZodType<any>): any {
+    const generator = new OpenAPIGenerator("/tmp", { title: "t" });
+    const result = (generator as any).processZodType(schema);
+    return JSON.parse(JSON.stringify(result));
+  }
+
+  it("converts primitive types", () => {
+    expect(processed(z.string())).toEqual({ type: "string" });
+    expect(processed(z.number())).toEqual({ type: "number" });
+    expect(processed(z.boolean())).toEqual({ type: "boolean" });
+  });
+
+  it("converts enums with their values", () => {
+    expect(processed(z.enum(["asc", "desc"]))).toEqual({
+      type: "string",
+      enum: ["asc", "desc"],
+    });
+  });
+
+  it("converts string length constraints", () => {
+    expect(processed(z.string().min(2).max(5))).toEqual({
+      type: "string",
+      minLength: 2,
+      maxLength: 5,
+    });
+  });
+
+  it("converts numeric range constraints", () => {
+    expect(processed(z.number().min(1).max(9))).toEqual({
+      type: "number",
+      minimum: 1,
+      maximum: 9,
+    });
+  });
+
+  it("converts the coerced number with a default from the docs example", () => {
+    const schema = processed(z.coerce.number().int().positive().default(20));
+    expect(schema).toMatchObject({ type: "number", default: 20 });
+  });
+
+  it("converts arrays without throwing", () => {
+    expect(processed(z.array(z.string()))).toEqual({
+      type: "array",
+      items: { type: "string" },
+    });
+  });
+
+  it("converts objects containing arrays", () => {
+    expect(processed(z.object({ name: z.string(), tags: z.array(z.string()) }))).toEqual({
+      type: "object",
+      properties: {
+        name: { type: "string" },
+        tags: { type: "array", items: { type: "string" } },
+      },
+      required: ["name", "tags"],
+    });
+  });
+
+  it("marks optional types as nullable", () => {
+    expect(processed(z.number().optional())).toEqual({
+      type: "number",
+      nullable: true,
+    });
+  });
+
+  it("keeps descriptions", () => {
+    expect(processed(z.number().describe("a count"))).toEqual({
+      type: "number",
+      description: "a count",
     });
   });
 });
--- a/packages/farm/src/openapi/generator.ts
+++ b/packages/farm/src/openapi/generator.ts
@@ -45,7 +45,15 @@ export class OpenAPIGenerator {
    * Get type from Zod type
    */
   private getTypeFromZodType(zodType: z.ZodType<any>): AllowedType {
-    const typeName = (zodType as any)._def?.typeName;
+    const def = (zodType as any)._def;
+    // Zod 4 stores a lowercase name on `_def.type`; Zod 3 stored `_def.typeName`.
+    // In Zod 3 `_def.type` can also be the element schema of an array, so only
+    // trust it when it is a string.
+    const v4Type = typeof def?.type === "string" ? def.type : undefined;
+    if (v4Type && allowedType.has(v4Type)) {
+      return v4Type as AllowedType;
+    }
+    const typeName = def?.typeName;
     if (typeName) {
       if (typeName === "ZodString") return "string";
       if (typeName === "ZodNumber") return "number";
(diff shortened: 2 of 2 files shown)


Kinfe123:
Superseded — the same fix landed on main via #431. Closing.

Issue #429, opened 2026-08-22
Core test suite fails on Windows


kira-1011:
## Summary

The `@farm.js/core` test suite fails on Windows: 44 tests across 14 files. CI runs
`ubuntu-latest` only, so none of it is visible there.

Most of it is test-side. Fixing the first cause unblocked builds that had never run on
Windows, which then exposed two source bugs, one platform limitation, and one runtime
path that does not complete.

## Test-side causes

**1. `fs.symlink(target, link, "dir")` requires elevation — 25 tests, 7 files.**
Fixtures link packages into a temp `node_modules`. On Windows `"dir"` needs
`SeCreateSymbolicLinkPrivilege`, so setup throws and every test in the file fails:

```
EPERM: operation not permitted, symlink
  'E:\...\packages\farm' -> 'E:\...\.tmp-production-middleware-kMJhS3\node_modules\@farm.js\core'
```

Junctions need no privilege, and `type` is ignored on non-Windows platforms, so one
unconditional `"junction"` is correct on every platform — 14 call sites, some with the
type on its own line. It also matches what a real install looks like on Windows: pnpm
links every package in `node_modules` as a junction, not a symlink.

Affected: `production-prebuilt-ssr` (20), `production-ssg` (2), `production-middleware`,
`production-observability`, `development-observability`, `production-public-dir`,
`fixtures/middleware-production-fixture`.

**2. Assertions hardcode POSIX absolute paths — 8 tests.**
`/test` and `/workspace` are not absolute on Windows, so `path.resolve` turns them into
`E:\test` while the expected value stays a POSIX literal:

```
expected 'E:\test\src\app\page.vue' to be '/test/src/app/page.vue'
expected { '@': 'E:\workspace\app\web' } to deeply equal { '@': '/workspace/app/web' }
```

The source is correct here: it keeps native paths internally and converts to POSIX only
at the client boundary (`toUrlPath` in `routing/route-manager.ts`). `utils.test.ts:30`
shows the inconsistency plainly — it calls `path.resolve()` for the input but asserts a
POSIX literal for the output, while real Vite on Windows emits `/@fs/E:/...`.

Affected: `route-manager` (4), `utils`, `create-server`, `i18n`, `docs-adapter`.

**3. An order-dependent asset lookup — 1 test.**
`production-middleware` selects a built asset with
`readdir(...).find((file) => file.startsWith("product-"))`. The client build emits that
[...]

PR #441, opened 2026-08-22
fix(core): key the PPR shell cache by locale

### LINKED ISSUE #435: PPR shell cache key has no locale, so one shell is shared across languages
## Summary

`ServerRenderer`'s PPR shell cache key has no locale in it, so one cached shell is shared by
every language. With `i18n.routing: "none"` a shell rendered in French is served to an English
request.

The docs already state the opposite, in `docs/src/app/docs/internationalization/page.md:298`:

> `routing: "none"` keeps localized pages dynamic because one static URL cannot safely contain
> multiple languages. PPR shell keys also include the locale.

## Cause

`packages/farm/src/server/renderer.ts:627` uses path and search only:

```ts
private getPPRCacheKey(pathname: string, search = ""): string {
  return createFarmCacheKey(["ppr", normalizeRevalidatePath(pathname), search]);
}
```

The production runtime does include it. `packages/farm/src/nitro/universal-build.ts:5379`:

```ts
function getPPRShellCacheKey(url, locale) {
  return createFarmCacheKey(["ppr", locale || "", normalizeRevalidatePath(url.pathname), url.search]);
}
```

The cached shell is locale-specific: `<html lang>`, `dir`, the serialized `window.__FARM_I18N__`
catalog and the `hreflang` links are all in it (`renderer.ts:1991`, `:2039`, `:2265`).

`PPRShellCacheOptions` (`renderer.ts:99`) carries no locale either, so neither
`cachePPRShell` (`:658`) nor `getCachedPPRShell` (`:654`) can pass one.

## Reproduction

```ts
import { describe, expect, it } from "vitest";
import { resolveFarmI18nConfig } from "../i18n/config";
import { FarmI18nRuntime } from "../i18n/runtime";
import { ServerRenderer } from "../server/renderer";

function createRenderer() {
  const i18n = resolveFarmI18nConfig(
    { locales: ["en", "fr"], defaultLocale: "en", routing: "none" },
    { root: "/test", mode: "production" },
  );
  const config = { root: "/test", srcDir: "src", outDir: "dist", basePath: "/", i18n };
  return new ServerRenderer(config as any, {} as any, new FarmI18nRuntime(i18n));
}

const request = (lang: string) =>
  new Request("http://localhost/dashboard", { headers: { "accept-language": lang } });

describe("ppr shell cache", () => {
  it("does not share a shell across locales", async () => {
    const renderer = createRenderer();
    const french = '<html lang="fr" dir="ltr">bonjour</html>';

    await renderer.runWithRequestContext(request("fr"), () =>
      (renderer as any).cachePPRShell({ pathname: "/dashboard", search: "" }, french),
    );

    const served = await renderer.runWithRequestContext(request("en"), async () =>
      (await (renderer as any).getCachedPPRShell("/dashboard", ""))?.value?.html,
    );

    expect(served).not.toBe(french);
  });
});
```

Fails:

```
FAIL  src/__tests__/ppr-shell-locale.test.ts > ppr shell cache > does not share a shell across locales
AssertionError: expected '<html lang="fr" dir="ltr">bonjour</ht…' not to be '<html lang="fr" dir="ltr">bonjour</ht…' // Object.is equality
```

Both requests produce the same key, `["ppr","/dashboard",""]`.

Four more checks I ran, all confirming the path is reachable:

- `getPPRShellBypassReason` returns `undefined` for a plain GET that differs only by
  `Accept-Language`, so nothing bypasses the cache.
- The response does get `Vary: ["Cookie", "Accept-Language"]`, so a CDN splits these correctly.
  Only Farm's own cache does not.
- With the default `prefix-except-default` routing the French request redirects to
  `/fr/dashboard`, so the paths differ and there is no collision. The bug needs
  `routing: "none"`.
- `["ppr","fr","/dashboard",""]` and `["ppr","en","/dashboard",""]` are distinct keys, so adding
  the locale is enough to separate them.

## Where it runs

`ServerRenderer` drives the dev server (`vite.ts`) and build-time SSG prerendering
(`build/index.ts`). A deployed app runs the generated Nitro runtime, which uses the locale-aware
key above, so this is a dev and build-time bug rather than a confirmed production one.

One thing I could not settle: `nitro/server-entry.ts` has a runtime slot for a `ServerRenderer`
(`registry.serverRenderer`, `undefined` by default at `:77`). If a deployed build ever registers
one, the same collision would apply in production.

## Suggested fix

Add `locale` to `PPRShellCacheOptions`, take it from `getFarmI18nClientSnapshot()?.locale ?? ""`
in the PPR block (`renderer.ts:1065`), and put it in the key in the same position production
uses.

Worth care: the key is built in three places. `getPPRCacheKey` is called from
`getCachedPPRShell`, `cachePPRShell` and `renderer.ts:1092`. If the read and write sides do not
pass the same locale the shell cache silently stops hitting instead of being fixed, so a test
should cover the round trip.

## Environment

- `@farm.js/core` `0.1.0-beta.56`
- Node.js 20.19.0, Windows 11

---

@Kinfe123 I'd like to take this one. Could you assign it to me?

surafel58:
Fixes #435

## What changed

In `packages/farm/src/server/renderer.ts`:

- `getPPRCacheKey` now takes a locale and puts it in the key, in the same position the
  generated Nitro runtime uses: `["ppr", locale, pathname, search]`.
- `PPRShellCacheOptions` carries a `locale`, resolved once from
  `getFarmI18nClientSnapshot()?.locale ?? ""` when the options are built.
- `getCachedPPRShell` takes the locale too, and the lookup in the render path reads it from the
  same `pprShellOptions` object that the later store uses.
- New tests in `packages/farm/src/__tests__/ppr-shell-locale.test.ts`.

## Why

The shell key was `["ppr", pathname, search]`, so one cached shell was shared by every language.
The shell is not language neutral: it holds `<html lang>`, `dir`, the serialized
`window.__FARM_I18N__` catalog and the `hreflang` links. With `i18n.routing: "none"` there is no
locale in the URL and no redirect, so two requests differing only by `Accept-Language` hit the
same entry and the second one gets the first one's language.

Nothing else stopped it. `getPPRShellBypassReason` bypasses on method, cookie, authorization,
refresh header and middleware or plugin context, but not on language. The response does get
`Vary: ["Cookie", "Accept-Language"]`, so a CDN splits these correctly; only Farm's own shell
cache did not.

`docs/src/app/docs/internationalization/page.md:298` already documents the intended behaviour:
"PPR shell keys also include the locale." The generated production runtime
(`nitro/universal-build.ts:5379`) already did this, so this also removes a difference between
the dev server and production.

## Why the locale is captured instead of read at write time

`cachePPRShell` is called from a streaming `onComplete` callback (`renderer.ts:1446`), which can
run outside the request context that resolved the locale. Reading the snapshot there could yield
`""` while the lookup used `"fr"`, and the two sides would then never agree, which would silently
stop the shell cache from hitting instead of fixing it. Resolving the locale once when
`pprShellOptions` is built and using that value for both sides avoids this by construction.

## Validation

The three tests that assert locale separation fail on `main` and pass with this change. Verified
by stashing only `renderer.ts` and keeping the test file:

```
without the fix: Tests  3 failed | 3 passed (6)
  FAIL does not serve a shell cached in one locale to another locale
  FAIL keeps a separate shell per locale for one path
  FAIL puts the locale in the key without dropping the path or search

with the fix:    Tests  6 passed (6)
```

The other three cover the round trip in one locale, separation by search params, and that the
locale a request resolves to is the one captured. They pass either way and are there to catch the
read and write sides drifting apart.

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/ppr-shell-locale.test.ts`: 6 passed
- `pnpm --filter @farm.js/core exec tsc --noEmit`: clean
- `pnpm exec oxlint` on both changed files: 0 warnings, 0 errors
- `pnpm exec oxfmt --check` on both changed files: clean
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm --filter @farm.js/core build`: ESM, CJS and DTS all succeed
- Full `@farm.js/core` suite run twice on this commit, once with the change and once with only
  `renderer.ts` reverted, then the failing test names compared. The only tests that fail without
  the change and pass with it are the three above. The only three that appear in the other
  direction are `app-markdown`, `layers` and `observability`, all `Test timed out in 5000ms`.
  Those are flaky under load rather than caused by this change: `app-markdown` and
  `observability` pass when run on their own, and `layers` fails the same way with
  `renderer.ts` reverted. Everything else failing is the pre-existing Windows set tracked in
  #429.

## Note

The key shape changes, so shells cached by an older build are no longer read and are rendered
once more after deploying. Nothing else reads that key.

---
## Summary by cubic
Keys the PPR shell cache by locale to prevent cross-language shells. Old behavior: key was ["ppr", pathname, search]; new behavior: ["ppr", locale, pathname, search], aligning the dev server with production and the docs.

**Changes**
- Resolve the locale once when building PPR shell options and reuse it for both lookup and store to avoid request-context drift during streaming completion.
- `getPPRCacheKey` and `getCachedPPRShell` now accept a locale; lookups and writes use the same locale.
- Add tests to verify per-locale separation and that path and search remain part of the key.

**Rollout**
- Key shape change invalidates prior entries; existing shells will miss and re-render once after deploy.
- No app-level config or API changes; CDN behavior unchanged (responses still vary on Accept-Language).

<sup>Written for commit fcd529e050e9057faa3ac876a5d93dd7833b54ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/ppr-shell-locale.test.ts
+++ b/packages/farm/src/__tests__/ppr-shell-locale.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it } from "vitest";
+import { resolveFarmI18nConfig } from "../i18n/config";
+import { FarmI18nRuntime } from "../i18n/runtime";
+import { getFarmI18nClientSnapshot } from "../i18n/server";
+import { ServerRenderer } from "../server/renderer";
+
+function createRenderer(routing: "none" | "prefix-except-default" = "none") {
+  const i18n = resolveFarmI18nConfig(
+    { locales: ["en", "fr"], defaultLocale: "en", routing },
+    { root: "/test", mode: "production" },
+  );
+  const config = { root: "/test", srcDir: "src", outDir: "dist", basePath: "/", i18n };
+  const renderer = new ServerRenderer(config as any, {} as any, new FarmI18nRuntime(i18n));
+  return renderer as any;
+}
+
+function request(locale: string, pathname: string) {
+  return new Request(`http://localhost${pathname}`, {
+    headers: { "accept-language": locale },
+  });
+}
+
+// The shell cache is a process-wide singleton, so each test uses its own path.
+describe("PPR shell cache and locale", () => {
+  it("does not serve a shell cached in one locale to another locale", async () => {
+    const renderer = createRenderer();
+    const french = '<html lang="fr" dir="ltr">bonjour</html>';
+
+    await renderer.cachePPRShell({ pathname: "/cross", search: "", locale: "fr" }, french);
+
+    expect(await renderer.getCachedPPRShell("/cross", "", "en")).toBeFalsy();
+  });
+
+  it("serves a cached shell back to the same locale", async () => {
+    const renderer = createRenderer();
+    const french = '<html lang="fr" dir="ltr">bonjour</html>';
+
+    await renderer.cachePPRShell({ pathname: "/same", search: "", locale: "fr" }, french);
+
+    // Guards the read and write sides against drifting apart, which would silently
+    // stop the shell cache from ever hitting.
+    expect((await renderer.getCachedPPRShell("/same", "", "fr"))?.value?.html).toBe(french);
+  });
+
+  it("keeps a separate shell per locale for one path", async () => {
+    const renderer = createRenderer();
+    const french = '<html lang="fr">bonjour</html>';
+    const english = '<html lang="en">hello</html>';
+
+    await renderer.cachePPRShell({ pathname: "/both", search: "", locale: "fr" }, french);
+    await renderer.cachePPRShell({ pathname: "/both", search: "", locale: "en" }, english);
+
+    expect((await renderer.getCachedPPRShell("/both", "", "fr"))?.value?.html).toBe(french);
+    expect((await renderer.getCachedPPRShell("/both", "", "en"))?.value?.html).toBe(english);
+  });
+
+  it("still separates entries by search params within one locale", async () => {
+    const renderer = createRenderer();
+    const first = "<html>page one</html>";
+
+    await renderer.cachePPRShell({ pathname: "/search", search: "?page=1", locale: "fr" }, first);
+
+    expect((await renderer.getCachedPPRShell("/search", "?page=1", "fr"))?.value?.html).toBe(first);
+    expect(await renderer.getCachedPPRShell("/search", "?page=2", "fr")).toBeFalsy();
+  });
+
+  it("captures the locale the request resolved to", async () => {
+    const renderer = createRenderer();
+
+    const resolved = async (locale: string) =>
+      renderer.runWithRequestContext(
+        request(locale, "/resolved"),
+        async () => getFarmI18nClientSnapshot()?.locale ?? "",
+      );
+
+    expect(await resolved("fr")).toBe("fr");
+    expect(await resolved("en")).toBe("en");
+  });
+
+  it("puts the locale in the key without dropping the path or search", async () => {
+    const renderer = createRenderer();
+
+    expect(renderer.getPPRCacheKey("/dashboard", "?a=1", "fr")).not.toBe(
+      renderer.getPPRCacheKey("/dashboard", "?a=1", "en"),
+    );
+    expect(renderer.getPPRCacheKey("/dashboard", "", "fr")).not.toBe(
+      renderer.getPPRCacheKey("/other", "", "fr"),
+    );
+    expect(renderer.getPPRCacheKey("/dashboard", "?a=1", "fr")).not.toBe(
+      renderer.getPPRCacheKey("/dashboard", "?a=2", "fr"),
+    );
+  });
+});
--- a/packages/farm/src/server/renderer.ts
+++ b/packages/farm/src/server/renderer.ts
@@ -99,6 +99,14 @@ interface CachedPPRShell {
 interface PPRShellCacheOptions {
   pathname: string;
   search: string;
+  /**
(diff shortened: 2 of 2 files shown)


Kinfe123 (commented):
ran this against c0de8dc8 before and after the change:

- the issue's repro fails on main as described, fr shell served to the en request under `["ppr","/dashboard",""]`
- on this branch all 6 tests pass. reverted only `renderer.ts` and got exactly the 3 locale failures with the other 3 still green, so the tests genuinely pin the fix and the round-trip
- key shape `["ppr", locale, path, search]` matches `getPPRShellCacheKey` in the nitro runtime, so dev/SSG and production now agree
- all four `getPPRCacheKey` call sites are covered, and the single construction point for `PPRShellCacheOptions` now carries the locale, so the read and write sides can't drift
- capturing the locale at options-build time instead of inside `getPPRCacheKey` is the right move given the streaming `onComplete` can run outside the request context. reading it there would have keyed writes on an empty locale and silently zeroed the hit rate

unit/e2e ci is green (the two vercel deploy checks failing are unrelated to this change). lgtm

Kinfe123 (approved):

Issue #440, opened 2026-08-22
Windows: remaining failures in auth, cli, and plugin


kira-1011:
## Summary

Three packages still fail on Windows after #429. They are skipped by
`scripts/test-packages.js` on `win32` so the rest of the suite can run on
`windows-latest`, and tracked here.

## `@farm.js/auth` (1 file)

```
FAIL  test/runtime.test.ts
Error: EBUSY: resource busy or locked, unlink 'C:\...\farm-auth-test-AQ8ZiI\.farm\auth.sqlite'
```

`afterAll` removes the temp directory while the runtime's `DatabaseSync` is still
open. `closeDatabase` in `src/runtime.ts` only closes the migration connection, and
the integration exposes no dispose, so the test has nothing to call first. Retrying
the `rm` would not help, since the handle is held for the life of the process.

## `@farm.js/cli` (4 tests, 77 passing)

```
✖ plans the generated Cloudflare Agent config before the first build
✖ applies only safe additive fixes
✖ recognizes Vue pages and creates a Vue root layout
✖ applies safe fixes inside a configured project root
```

The first compares a rendered command string. `formatCommandArgument` in
`src/deploy.ts` treats a backslash as unsafe, so every Windows path is wrapped in
POSIX single quotes. Execution is unaffected because it uses `executable` plus
`args`, but the displayed command is not valid on Windows shells. The other three
are separate and not yet diagnosed.

## `@farm.js/plugin` (3 tests, 41 passing)

```
FAIL  src/rsc/core-external.test.ts > does not prefix absolute Vite environment output paths with the project root
  → expected '\workspace\app\.nitro\vite\dist\rsc\index.js' to be '/workspace/app/.nitro/vite/dist/rsc/index.js'

FAIL  src/rsc/core-external.test.ts > builds and boots a real RSC app with exact-root runtime imports outside the workspace
FAIL  src/rsc/nitro-build.test.ts > boots node-server output after the project and workspace dependencies are removed
  → C:\...\dist\rsc\index.js is resolved as a module now, but it was an external
    module before. Please check whether there are conflicts in your Rollup options
    "external" and "manualChunks"
```

The first is a POSIX path literal, the same class as #429. The other two are a
Rollup externals conflict with Windows absolute paths and need investigation.

## Environment

- `@farm.js/core` `0.1.0-beta.56`
- Windows 11, Node.js 24, non-elevated shell

---

@Kinfe123, happy to pick these up once #429 lands. Could you assign this to me?

[...]

Issue #434, opened 2026-08-22
Windows: production middleware test hangs in fixture cleanup


kira-1011:
## Summary

`production-middleware.test.ts` times out after 120s on Windows. The time goes to fixture
teardown, not to the test itself.

## Cause

The build copies sharp's native binary into the output, and the test process loads it while
optimizing images. Windows will not unlink a loaded module:

```
EPERM: operation not permitted, unlink
  .vercel/output/functions/__nitro.func/node_modules/@img/sharp-win32-x64/lib/sharp-win32-x64.node
```

`cleanupMiddlewareProductionFixture` uses a recursive `fs.rm` with retries, so that one file is
retried at every parent directory on the way up and the remove never finishes. The same
directory deletes in 23ms once the process has exited.

## Evidence

Every request completes quickly, including the one that loads the wasm image codec:

```
[434] <- #2  GET /dashboard/opengraph-image in 3ms
[434] <- #24 GET /rewrite-alias in 3ms
```

All 24 requests finish and the whole test body runs in under 9 seconds.

## Correction to the original report

This was first filed as a stall in the wasm image codec. That trace was the last thing logged
before the hang, but the hang happens about twenty requests later, in teardown. The codec is
not involved.

## Environment

- Windows 11, Node.js 24
- Reproduced from beta.59 through beta.67

Kinfe123:
can't reproduce this on macos/linux (test passes here), so taking your windows trace as the source of truth, and it's a good one: the `AGFzbQ` base64 fetch means the bundled vercel output is loading the wasm codec inline even though `@img/sharp-win32-x64` is sitting right there in the output. so there are really two questions:

1. why the native binding isn't picked up on windows (path/resolution in the bundled output would be my first suspect, given how many windows path bugs #429 already surfaced in that area)
2. why the wasm fallback stalls instead of finishing slow

if the native binding resolution turns out to be the fix, the stall may never need solving. worth logging what `sharp`'s loader actually throws when it falls back, that usually names the path it tried.

kira-1011:
Small correction before either of us digs into sharp: the wasm is not sharp's.

It comes from `@vercel/og`, which we depend on directly (`packages/farm/package.json:598`)
and call in `metadata-image.ts:203` for OpenGraph images. `@vercel/og@0.11.1` pulls
`@resvg/resvg-wasm@2.4.0` for SVG rasterising, and that library is wasm by design. There is
no native resvg in the tree.

[...]

Issue #428, opened 2026-08-22
storage.dispose() leaves db0 database connections open


kira-1011:
## Summary

`loadDatabaseDriver` creates a `db0` database and passes it to unstorage's `db0`
driver, but nothing ever closes it. `unstorage/drivers/db0` does not implement
`dispose`, so `storage.dispose()` and `disposeStorage()` resolve while the
connection stays open for the lifetime of the process.

This affects every database-backed driver resolved through `DATABASE_CONNECTORS`:
`sqlite`, `postgres`, `mysql`, `pglite`, and `libsql`.

## Impact

- **SQL backends** (`postgres`, `mysql`, `libsql`): a connection pool is leaked per
  disposed storage instance. Repeated create/dispose cycles — dev server restarts,
  test suites, serverless invocations — accumulate connections until the server's
  limit is reached.
- **SQLite**: the database file stays locked. On Windows the file cannot be removed
  or renamed after disposal, which fails 8 tests in `packages/farm/src/__tests__/storage.test.ts`
  with `EBUSY: resource busy or locked, unlink '...\app.sqlite'`.

The leak is platform independent; Windows only makes it observable, because it
refuses to unlink a file that still has an open handle.

## Reproduction

`packages/farm/src/__tests__/storage-dispose.test.ts`:

```
× closes a database it created from a config
  → expected "dispose" to be called at least once
✓ leaves a caller-provided database open
```

<details>
<summary>Test file</summary>

```ts
import { mkdtemp, rm } from "node:fs/promises";
import { tmpdir } from "node:os";
import path from "node:path";
import { afterEach, describe, expect, it, vi } from "vitest";
import { createFarmStorage, databaseStorage, sqliteStorage } from "../storage";

const { created } = vi.hoisted(() => ({ created: [] as any[] }));

/** Spies on every db0 database while leaving the real implementation in place. */
vi.mock("db0", async (importOriginal) => {
  const actual = await importOriginal<typeof import("db0")>();
  return {
    ...actual,
    createDatabase: (connector: any) => {
      const database = actual.createDatabase(connector);
      vi.spyOn(database, "dispose");
      created.push(database);
      return database;
    },
  };
});

const [nodeMajor, nodeMinor] = process.versions.node.split(".").map(Number);
const supportsNodeSqlite = nodeMajor > 22 || (nodeMajor === 22 && nodeMinor >= 5);

const tempDirs = new Set<string>();

async function createTempDir(prefix: string): Promise<string> {
[...]

PR #448, opened 2026-08-22
fix(core): make the test suite pass on Windows

### LINKED ISSUE #429: Core test suite fails on Windows

kira-1011:
## Summary

Fixes #429.

The suite had 44 failing tests across 14 files on Windows. None of it was visible to CI,
which only ran `ubuntu-latest`. Most were platform assumptions in the tests, but three
were real defects.

**`resolveModuleSourcePath` could not decode Windows `/@fs/` ids.** It stripped four
characters from `/@fs/C:/...`, leaving `/C:/...`, which resolves to nothing. Vite emits
exactly that id shape on Windows, so `"use client"` detection failed for any module
referenced through it.

**`sanitizeStack` left native separators** while `displayPath` in the same module
normalizes to `/`, so one diagnostics object reported two different path shapes.

**Vitest's threads pool segfaults during teardown on Windows.** Every test passed, the
summary printed, then the process exited 139. CI would have failed on a green suite.

## Changes

- Use `"junction"` for the fixture symlinks, 14 sites in `packages/farm` and one in
  `@farm.js/eve`. Windows grants `"dir"` symlinks only to elevated processes, junctions
  need no privilege, and the type argument is ignored on other platforms. pnpm links
  packages the same way on Windows.
- Build expected paths with `path.join` and `path.resolve` instead of POSIX literals in
  `route-manager`, `utils`, `create-server`, `i18n`, and `docs-adapter`. The source keeps
  native paths internally and converts at the client boundary, so the assertions were the
  platform-specific half.
- `production-middleware` asserts that the emitted asset set contains the name the HTML
  references. The client build emits that image under two hashed names, so `readdir` order
  decided the result.
- Add `maxRetries` and `retryDelay` to the temp cleanups that race Windows handle release.
- Decode `/@fs/` ids the way Vite's own `fsPathFromId` does, and normalize the result to a
  native path so it matches the function's other branches.
- Normalize separators inside the `<project>` token in `sanitizeStack`.
- Skip the four SIGTERM assertions on Windows. `child.kill("SIGTERM")` maps to
  `TerminateProcess`, so graceful shutdown cannot be triggered from a parent process and
  its effects are not observable.
- Run the forks pool sequentially on Windows in `@farm.js/core` and `@farm.js/react`.
  Unbounded forks hit `listen ENOBUFS` in the fixtures that bind real servers.
- Add one `windows-latest` entry on Node 24 to the CI matrix.

This also carries the `loadDatabaseDriver` disposal from #428, without which
`storage.test.ts` fails eight tests on Windows and the new job cannot pass. The
`databaseStorage` ownership question in that issue is unaffected and still open.

## Test plan

Windows 11, Node.js 24, non-elevated shell.

- [x] `node scripts/test-packages.js --skip-incompatible`, the command the new entry runs:
      `Tested 28 packages. Skipped 3 packages.` exit 0
- [x] `packages/farm`: 143 files, 1147 passed, 6 skipped, exit 0
- [x] `@farm.js/react`: 24 files, 191 passed, exit 0
- [x] `@farm.js/eve`: 3 passed
- [x] `pnpm --filter @farm.js/core type-check`
- [x] Root `pnpm lint`, including the new synthetic-popstate check
- [x] Confirmed the 44 failures on `main` at `58ad3198` before the fix, and again on
      `0.1.0-beta.56`

## Deferred

- `@farm.js/auth`, `@farm.js/cli`, and `@farm.js/plugin` still fail on Windows.
  `scripts/test-packages.js` skips them on `win32` and logs each one. Tracked in #440.
- The image optimization test is skipped on Windows pending #434.

One thing worth flagging separately: `@farm.js/core` declares no `engines` field, so the
existing Node 22 and 24 entries skip it through `--only-with-node-engine`. Today it is
covered only by the Node 20 entry. The new Windows entry omits that flag for the same
reason.

---
## Summary by cubic
Make the test suite pass on Windows and run it in CI. Fix path handling and database disposal that broke Vite resolution and left SQLite files locked.

- CI: add a `windows-latest`/Node 24 matrix entry without `--only-with-node-engine`; keep the generated-types check on `ubuntu-latest`/Node 20; on Windows set `TEMP`/`TMP` to `${{ runner.temp }}` to avoid percent-encoded short paths.
- Repo: enforce LF checkouts via `.gitattributes` so source-snippet tests pass on Windows; raise the docs git-history test timeout to accommodate slower spawns on the runner.
- Runtime: `resolveModuleSourcePath` decodes and normalizes Windows `/@fs/` ids; `sanitizeStack` emits POSIX separators within the `<project>` token; `loadDatabaseDriver` now disposes the `db0` database it creates (tests verify created databases close and caller-provided databases stay open); the Cloudflare agent test imports the generated worker from an unencoded file URL.
- Windows-only/test robustness: use junction symlinks; build expected paths with `path.join`/`path.resolve`; assert image assets by name set instead of directory order; add rm retries for delayed handle release; run Vitest with a single-process forks pool in `@farm.js/core` and `@farm.js/react`; skip SIGTERM-based shutdown assertions and the image optimization test on Windows; `scripts/test-packages.js` skips `@farm.js/auth`, `@farm.js/cli`, and `@farm.js/plugin` on Windows.

<sup>Written for commit 8ac7d7b2fecfcb73cd3cae93f2e1097b1ab07f79. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/448?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Check out with LF everywhere. Several tests match multi-line snippets of source
+# files, and a CRLF checkout on Windows breaks those comparisons.
+* text=auto eol=lf
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,19 +9,27 @@ on:
 
 jobs:
   test:
-    name: Test
-    runs-on: ubuntu-latest
+    name: Test (${{ matrix.os }}, Node ${{ matrix.node-version }})
+    runs-on: ${{ matrix.os }}
 
     strategy:
       fail-fast: false
       matrix:
         include:
-          - node-version: 20
+          - os: ubuntu-latest
+            node-version: 20
             test_args: --skip-incompatible
-          - node-version: 22
+          - os: ubuntu-latest
+            node-version: 22
             test_args: --skip-incompatible --only-with-node-engine
-          - node-version: 24
+          - os: ubuntu-latest
+            node-version: 24
             test_args: --skip-incompatible --only-with-node-engine
+          # No --only-with-node-engine: @farm.js/core declares no engines field and
+          # would otherwise be skipped, which is the suite this entry exists to cover.
+          - os: windows-latest
+            node-version: 24
+            test_args: --skip-incompatible
 
     steps:
       - name: Checkout
@@ -57,7 +65,7 @@ jobs:
           NODE_OPTIONS: --max-old-space-size=8192
 
       - name: Check committed generated types
-        if: matrix.node-version == 20
+        if: matrix.os == 'ubuntu-latest' && matrix.node-version == 20
         run: pnpm --dir docs exec farm generate --check
 
       - name: Run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,11 @@ jobs:
 
       - name: Run tests
         run: pnpm test -- ${{ matrix.test_args }}
+        env:
+          # The Windows runner points TEMP at the short name C:\Users\RUNNER~1, and
+          # Vite cannot resolve a dynamic import whose path it percent-encodes.
+          TEMP: ${{ runner.temp }}
+          TMP: ${{ runner.temp }}
 
       - name: Type check
         run: pnpm type-check
--- a/packages/farm-cf-agent/src/__tests__/cf-agent.test.ts
+++ b/packages/farm-cf-agent/src/__tests__/cf-agent.test.ts
@@ -108,7 +108,10 @@ export default { fetch() { return new Response("agent"); } };
     expect(generated.assets.directory).toBe("./.output/public");
     expect(generated.durable_objects.bindings[0].class_name).toBe("CounterAgent");
 
-    const module = await import(`${pathToFileURL(result.wrapperPath).href}?test=${Date.now()}`);
+    // decodeURI because Vite resolves this id itself and fails on a percent-encoded
+    // path, which a Windows short name such as RUNNER~1 produces.
+    const wrapperUrl = decodeURI(pathToFileURL(result.wrapperPath).href);
+    const module = await import(`${wrapperUrl}?test=${Date.now()}`);
     expect(
       module.default.fetch(new Request("https://example.com/agents/counter/default")),
     ).toMatchObject({ status: 200 });
--- a/packages/farm-eve/src/__tests__/eve.test.ts
+++ b/packages/farm-eve/src/__tests__/eve.test.ts
@@ -47,7 +47,8 @@ describe("Eve Vercel output", () => {
     await writeFile(join(root, "package.json"), '{"type":"module"}\n');
     await writeFile(join(workspaceRoot, ".vercel", "project.json"), "{}\n");
     await writeFile(join(outputDirectory, "config.json"), '{"version":3,"routes":[]}\n');
-    await symlink(packageDirectory, join(root, "node_modules", "eve"), "dir");
+    // Junctions need no Windows privilege; the type is ignored on POSIX.
+    await symlink(packageDirectory, join(root, "node_modules", "eve"), "junction");
 
     await writeEveVercelOutput({
       root,
--- a/packages/farm-react/vitest.config.ts
+++ b/packages/farm-react/vitest.config.ts
@@ -9,6 +9,11 @@ export default defineConfig({
     },
   },
   test: {
+    // Windows: the threads pool segfaults during teardown, so the run exits
+    // non-zero with every test passing.
+    ...(process.platform === "win32"
+      ? { pool: "forks" as const, poolOptions: { forks: { minForks: 1, maxForks: 1 } } }
+      : {}),
     environment: "jsdom",
     globals: true,
     include: ["src/**/*.test.{ts,tsx}"],
--- a/packages/farm/src/__tests__/client-component.test.ts
+++ b/packages/farm/src/__tests__/client-component.test.ts
@@ -1,3 +1,4 @@
+import { pathToFileURL } from "node:url";
 import fs from "fs";
 import os from "os";
 import path from "path";
@@ -83,10 +84,13 @@ describe("client component path resolution", () => {
       "'use client';\nexport default function Page() { return null; }\n",
     );
 
-    expect(resolveModuleSourcePath(`file://${sourceFile}`, root)).toBe(sourceFile);
-    expect(isClientComponentModule(`file://${sourceFile}`, root)).toBe(true);
-    expect(resolveModuleSourcePath(`/@fs${sourceFile}`, root)).toBe(sourceFile);
-    expect(isClientComponentModule(`/@fs${sourceFile}`, root)).toBe(true);
+    const fileUrl = pathToFileURL(sourceFile).href;
+    const fsId = `/@fs/${sourceFile.split(path.sep).join("/").replace(/^\/+/, "")}`;
+
+    expect(resolveModuleSourcePath(fileUrl, root)).toBe(sourceFile);
+    expect(isClientComponentModule(fileUrl, root)).toBe(true);
+    expect(resolveModuleSourcePath(fsId, root)).toBe(sourceFile);
+    expect(isClientComponentModule(fsId, root)).toBe(true);
   });
 
   it("supports hydratable server pages through export const hydrate = true", () => {
--- a/packages/farm/src/__tests__/create-server.test.ts
+++ b/packages/farm/src/__tests__/create-server.test.ts
@@ -1,3 +1,4 @@
+import path from "node:path";
 import { describe, expect, it } from "vitest";
 import {
   createFarmClientOptimizeDepsConfig,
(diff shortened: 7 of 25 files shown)


Kinfe123:
windows failure was a pre-existing cf-agent bug the new runner just exposed (the runner's temp path has `RUNNER~1` in it, `pathToFileURL` encodes the `~` to `%7E` and vitest never decodes it), not your changes. saw you already pushed the decodeURI fix, ran the cf-agent tests on your branch and all 5 pass here. once the windows job goes green this is good to merge.

Kinfe123 (approved):

kira-1011:
@Kinfe123 PR is ready

Kinfe123 (approved):

kira-1011:
One note on the `pool: "forks"` override, since it should not stay in the tree forever.

It is a workaround for a vitest 1.x bug, not a Windows fact. I tested this locally on a
scratch branch, with the override removed and the default threads pool:

| vitest | result |
| --- | --- |
| 1.6.1 (current) | segfault at teardown, exit 139 |
| 3.2.7 | no segfault. `@farm.js/react` 24 files, 191 tests, exit 0. `@farm.js/core` 141 of 143 files, 1145 tests pass |
| 4.1.11 (latest) | does not start |

That is why Vite runs its own Windows job on the default pool: it is on vitest 4, not
because its suite is gentler.

Two caveats on upgrading:

**3.2.7 is not drop-in.** Core exits 1 on two behaviour changes, unrelated to Windows:

```
FAIL  app-markdown.test.ts > renders MDX with configured components
FAIL  type-artifacts.test.ts > types global CSS side effects and CSS Modules
```

**4.1.11 is blocked on vite.** It needs `vite ^6 || ^7 || ^8`, while `packages/farm`
declares `^5.0.10` and resolves 5.4.20, so it fails at startup:

```
ERR_PACKAGE_PATH_NOT_EXPORTED: './module-runner' is not defined by "exports" in vite@5.4.20
```

So the follow-up is: upgrade to 3.2.7, migrate those two tests, and delete the override
from `packages/farm/vitest.config.ts` and `packages/farm-react/vitest.config.ts`. Latest
waits on a vite 6 bump, which is a framework dependency rather than test tooling.

Twelve packages declare vitest, so that is its own PR. Happy to take it if you want it
filed. Nothing else in this PR depends on the pool setting: the junctions, path
assertions, two source bugs, SIGTERM skips, `.gitattributes`, and the TEMP redirect all
stand on their own.

Kinfe123:
Would love you to take a look at as follow up.

Issue #458, opened 2026-08-22
Windows: RSC Nitro build hits EMFILE reading the pnpm store


kira-1011:
## Summary

On Windows, the RSC Nitro build exhausts file handles while reading the pnpm store, so
`core-external.test.ts` fails with `EMFILE`. Split out of #440, where the rest of
`@farm.js/plugin` now passes.

## Reproduction

`packages/farmjs-plugin/src/rsc/core-external.test.ts` on Windows 11, Node.js 24:

```
[FARM] Building RSC server with Nitro (preset: node-server)...
[error] [nitro] Error: EMFILE: too many open files, open
  'E:\.pnpm-store\v11\files\f9\8bd4b23d2a549ad9fdd53203af03adf173b624313dcb37ca82ff4382fda9...'

FAIL  src/rsc/core-external.test.ts > RSC core runtime bundling > builds and boots a real
      RSC app with exact-root runtime imports outside the workspace
Tests  1 failed | 4 passed (5)
```

The failure is inside `buildRscNitro`, in Nitro's own tracing of dependencies, rather than
in Farm code. `nitro-build.test.ts` runs the same build and passes, so this is about how
many files this fixture's dependency graph opens at once.

## Notes

- Passes on Linux. Both jobs in #456 are green with the package unskipped there.
- Windows applies a per-process limit to concurrent C runtime file descriptors, and Node
  exposes no way to raise it, so this may vary with the store layout and the machine.
  Worth confirming on a GitHub runner before assuming it reproduces everywhere.
- Nitro also logs a non-fatal `EPERM` for its `last-build` symlink during the same build,
  which is a separate Windows papercut and does not stop the build.

## Environment

- `@farm.js/core` `0.1.0-beta.56`
- Windows 11, Node.js 24, non-elevated shell
- Preset: `node-server`

---

@Kinfe123, happy to take this one too if you think it is worth chasing upstream. Could you
assign it to me?

Kinfe123:
makes sense. i think you should confirm it repros on the github runner first since windows file limits vary per machine. if it's nitro's tracing opening too many files at once then yeah it belongs upstream, worth filing there with the trace. fine keeping the test skipped on win32 until then.

Kinfe123:
[...]

PR #456, opened 2026-08-22
chore: upgrade vitest to 3.2.7

kira-1011:
## Summary

Closes #455.

#448 added `pool: "forks"` with a single fork on Windows for `@farm.js/core` and
`@farm.js/react`. vitest 1.x crashes during teardown there: every test passes, the summary
prints, then the run exits 139 or 127, so CI fails on a green suite. `@farm.js/plugin` hits
the same crash once its tests are no longer skipped.

vitest 3.2.7 does not crash. Measured with the overrides removed and the default pool:

| package | vitest 1.x | vitest 3.2.7 |
| --- | --- | --- |
| `@farm.js/react` | exit 127, no summary | 24 files, 191 tests, exit 0 |
| `@farm.js/core` | exit 139 | 145 files, 1172 tests, exit 0 |
| `@farm.js/plugin` | crash after 3 or 4 files | run completes, only the #440 failures remain |

Vite's own repo runs its Windows job on the default pool for the same reason.

## Changes

- Move the twelve packages that declare `vitest` from `^1.1.0` / `^1.6.1` to `^3.2.7`, and
  `@vitest/coverage-v8` with them since it is version locked.
- Delete the Windows pool override from `packages/farm/vitest.config.ts` and
  `packages/farm-react/vitest.config.ts`.
- Add explicit timeouts to `@farm.js/core` and `@farm.js/eve`, scaled for CI the way
  `vite/vitest.config.e2e.ts` does.

## Why the timeouts

The override removed all parallelism, which hid how tight the 5s default was. Restoring it
means the fixtures that bundle a config or spawn a subprocess compete for the machine.
Alone they are quick:

```
app-markdown.test.ts    635ms
type-artifacts.test.ts  2072ms
layers.test.ts          2250ms
```

Under a full 145 file run they pass 5s and fail, and which ones fail varies per run.

## Side effects worth noting

- `@farm.js/core` drops from 203s to 80s, since the override serialised it.
- The `EMFILE: too many open files` in `@farm.js/plugin` is gone. It only appeared under
  the single-fork override, so the workaround was causing it.
- `@farm.js/plugin` now runs to completion, which unblocks verifying the fixes in #440.

## Test plan

Windows 11, Node.js 24.

- [x] `node scripts/test-packages.js --skip-incompatible`: `Tested 28 packages. Skipped 3 packages.` exit 0
- [x] `@farm.js/core`: 145 files, 1172 passed, 6 skipped, exit 0
[...]

PR #464, opened 2026-08-22
fix(windows): run auth, cli, and plugin tests again

### LINKED ISSUE #440: Windows: remaining failures in auth, cli, and plugin

kira-1011:
## Summary

Closes #440.

The three packages #448 skipped on `win32` now run again, so
`scripts/test-packages.js` no longer needs its skip list. Each had a real bug behind it.

**`@farm.js/auth` never closed the database it opened.** The better-auth runtime is cached
on `globalThis`, but the reference to its connection was dropped, so nothing could close
it. On SQLite that held a file lock and failed the temp cleanup with `EBUSY`; on every
other driver it left a pool open for the life of the process.

**`@farm.js/cli` single quoted Windows paths in the commands it prints.**
`formatCommandArgument` treated a backslash as unsafe, so `--config C:\...\wrangler.jsonc`
was displayed wrapped in POSIX quotes, which is not valid in a Windows shell. Separately,
`doctor` reported project relative paths with native separators where the rest of the
package already uses `toPosix`.

**`@farm.js/plugin` matched externals with forward slash substrings.**
`id.includes("dist/rsc")` never matches `C:\...\dist\rsc\index.js`, so the same module was
external in one pass and not in the next, and Rollup rejected it with `is resolved as a
module now, but it was an external module before`.

## Changes

- `disposeFarmAuth()` closes the connection and drops the cached runtime, exported from
  `internal.ts` beside `migrateFarmAuth`. A reconfigure closes the connection the previous
  runtime was using. Execution paths that hand out a database are untouched.
- `formatCommandArgument` accepts a backslash on `win32` only, keeping the quote style, so
  untrusted values are still quoted. `doctor` reports paths through `toPosix`.
- The externals predicate normalises the id before matching.
- Both plugin build fixtures waited unbounded for a child exit after `SIGKILL`. Windows
  maps both signals to `TerminateProcess`, so an exit event that had already fired left the
  run hanging rather than failing. Bounded to 5s.
- `@farm.js/auth` gets the timeouts `@farm.js/core` and `@farm.js/eve` took in #456.
- Remove `WINDOWS_SKIPPED` from `scripts/test-packages.js`.

## Test plan

Windows 11, Node.js 24, on top of #456.

- [x] `node scripts/test-packages.js --skip-incompatible`: `Tested 31 packages. Skipped 0 packages.` exit 0
- [x] `@farm.js/auth`: 2 files, 6 tests
- [x] `@farm.js/cli`: 81 passed
- [x] `@farm.js/plugin`: 10 files, 43 passed, 1 skipped
- [x] `pnpm --filter @farm.js/core type-check`
- [x] Root `pnpm lint`

The auth dispose was written test first: the new test failed with
`disposeFarmAuth is not a function` before the implementation existed.

## Notes

One fixture stays skipped on Windows: `core-external.test.ts` builds an RSC app whose
dependency graph makes Nitro exhaust file handles reading the pnpm store, reported in #458.
Windows caps concurrent C runtime descriptors and Node cannot raise it, so that one may be
machine dependent and is worth confirming on a runner.

Fixing the unbounded teardown is what made this diagnosable. Both tests previously ran for
400 seconds without completing, which read as a hung build. The build had finished and the
assertions had passed; only the teardown was stuck.

---
## Summary by cubic
Re-enables Windows tests for `@farm.js/auth`, `@farm.js/cli`, and `@farm.js/plugin` by fixing resource cleanup, path quoting, and external matching. Handles dispose/reconfigure races in the auth runtime and removes the Windows skip list so these packages run on Windows again.

- `@farm.js/auth`: Keep and close the database handle; export `disposeFarmAuth()`; close on dispose and on reconfigure; wait for in-flight runtime creation to register before closing; tests ensure temp files self-clean.
- `@farm.js/cli`: Treat backslashes as safe on win32 so printed commands aren’t POSIX-quoted; `doctor` reports POSIX paths consistently.
- `@farm.js/plugin`: Normalize ids before external matching to avoid Windows inconsistencies; bound child-process teardown to 5s and fail if the child does not exit; one RSC build fixture remains skipped on Windows pending #458.
- `scripts/test-packages.js`: Remove the Windows skip list.
- `vitest`: Share higher test/hook timeouts via `vitest.shared.ts` and apply in multiple packages.

**Migration**
- If tests create an `@farm.js/auth` runtime, call `disposeFarmAuth()` in teardown to release SQLite file handles.

<sup>Written for commit 3fdd98d73f470cb109d568d941398a656708f029. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/464?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm-auth/src/internal.ts
+++ b/packages/farm-auth/src/internal.ts
@@ -1,6 +1,7 @@
 import type { ResolvedFarmAuthConfig } from "./types.js";
 import {
   configureFarmAuth,
+  disposeFarmAuth as runDispose,
   getFarmAuthRuntime,
   migrateFarmAuth as runMigration,
 } from "./runtime.js";
@@ -48,4 +49,8 @@ export function createFarmAuthIntegration(
 
 export async function migrateFarmAuth(): Promise<void> {
   await runMigration();
+}
+
+export async function disposeFarmAuth(): Promise<void> {
+  await runDispose();
 }
--- a/packages/farm-auth/src/runtime.ts
+++ b/packages/farm-auth/src/runtime.ts
@@ -23,6 +23,8 @@ interface RuntimeState {
   config?: ResolvedFarmAuthConfig;
   options?: FarmAuthRuntimeOptions;
   runtime?: Promise<BetterAuthRuntime>;
+  /** Kept so the connection opened for the runtime can be closed again. */
+  database?: unknown;
 }
 
 type GlobalAuthState = typeof globalThis & {
@@ -49,7 +51,10 @@ export function configureFarmAuth(
   state.config = config;
   state.options = options;
   if (changed) {
+    const previous = state.database;
     state.runtime = undefined;
+    state.database = undefined;
+    if (previous) void closeDatabase(previous).catch(() => {});
   }
 }
 
@@ -77,6 +82,21 @@ export function getFarmAuthRequest(request?: Request): Request {
   return current;
 }
 
+/**
+ * Close the connection opened for the runtime and drop the cached instance.
+ *
+ * The runtime is cached on globalThis and its connection stays open for the life
+ * of the process, which leaves a file lock behind on SQLite and a pool open on
+ * every other driver. Call this from a teardown hook.
+ */
+export async function disposeFarmAuth(): Promise<void> {
+  const state = getState();
+  const database = state.database;
+  state.runtime = undefined;
+  state.database = undefined;
+  await closeDatabase(database);
+}
+
 export async function migrateFarmAuth(): Promise<void> {
   const state = getState();
   if (!state.config?.enabled || !state.options) {
@@ -100,6 +120,7 @@ async function createRuntime(
   options: FarmAuthRuntimeOptions,
 ): Promise<BetterAuthRuntime> {
   const authOptions = await createBetterAuthOptions(config, options);
+  getState().database = authOptions.database;
 
   if (options.mode === "development" && config.database.migrateInDevelopment) {
     const { getMigrations } = await import("better-auth/db/migration");
--- a/packages/farm-auth/src/runtime.ts
+++ b/packages/farm-auth/src/runtime.ts
@@ -66,7 +66,18 @@ export async function getFarmAuthRuntime(): Promise<BetterAuthRuntime> {
     );
   }
 
-  state.runtime ||= createRuntime(state.config, state.options);
+  if (!state.runtime) {
+    const runtime = createRuntime(state.config, state.options, (database) => {
+      // A dispose or reconfigure that raced this creation already closed what it
+      // knew about, so an orphaned connection closes itself instead of leaking.
+      if (state.runtime === runtime) {
+        state.database = database;
+      } else {
+        void closeDatabase(database).catch(() => {});
+      }
+    });
+    state.runtime = runtime;
+  }
   return state.runtime;
 }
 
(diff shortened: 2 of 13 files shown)


kira-1011:
Pushed two changes since the first review pass.

**The RSC build fixture is no longer skipped.** #468 unskipped it against a runner and the
Windows job passed, so the skip was hiding a working test:

```
✓ src/rsc/core-external.test.ts (5 tests) 17825ms
  ✓ builds and boots a real RSC app with exact-root runtime imports outside the workspace  17752ms
```

#458 is closed as a local file descriptor limit rather than a repo problem. Worth being
explicit that it still fails on my machine, so anyone with a low limit may see it; the note
on that issue records where to start if it turns out to be common.

**Timeouts moved to a shared `vitest.shared.ts`** instead of a hand-written pair in three
package configs. The value also changed: Windows now gets the same headroom as CI, since
that is where the slowness is, and a Windows developer previously got the shorter one.
`nuxt` and `next.js` both branch on platform for this.

Verified locally: 30 packages pass, plus the 12 after `@farm.js/plugin` run explicitly since
the runner stops at the first failure. `@farm.js/plugin` is 43 of 44 here, the exception
being the `EMFILE` above, which CI has already shown passing. `type-check` and root `lint`
are clean.

Kinfe123 (approved):

kira-1011:
@Kinfe123  PR is ready

PR #516, opened 2026-08-24
refactor: share decodeRouteSegment across route matching

kira-1011:
Follow-up to #515, as offered there.

`decodeRouteSegment` had been copied into four places, each by the fix that needed it and
before a shared home existed: `integrations.ts` in #487, then the two middleware runtimes and
the emitted matcher in #508. #515 added `utils/decode` rather than a fifth copy.

This imports the shared helper in the three that can. The bodies were byte identical, so
there is no behaviour change.

The copies in `nitro/universal-build.ts` and `workflows.ts` stay inlined. Both are emitted
into generated runtime source that runs in its own bundle and cannot import from core.

## Verification

Core suite green, 1248 passed and 6 skipped, the same counts as before the change. The suites
covering the touched paths pass in isolation too, 212 tests across middleware, integrations,
router and workflows. Type check and lint clean.

## Note

Stacked on #515, so it shows that commit until #515 lands.

While checking I swept every `decodeURIComponent` in `packages/*/src`, 37 sites. Aside from
these, they are inline `try`/`catch` at the call site with their own fallback semantics, so I
left them alone. The one bare decode left is `farm-cli/src/explain.ts:615`, which reads a CLI
argument rather than a request path, so a malformed value there fails the command rather than
a response. Happy to file it separately if you think it is worth tidying.

---
## Summary by cubic
Deduplicates route-segment decoding and fixes workflow ID decoding errors. Previously, malformed percent-encodings in workflow IDs (e.g., caf%E9, %ZZ) threw and returned 500; they now fall back to the raw segment and return 404.

- Introduces a shared `decodeRouteSegment` in `packages/farm/src/utils/decode.ts` and replaces duplicated implementations in `integrations.ts`, `middleware/manager.ts`, and `middleware/production-runtime.ts` (no behavior change).
- Updates the workflows request handler to use the shared helper; adds an inlined copy in the emitted Nitro handler because generated runtime code cannot import from core.
- Adds tests verifying 404 on malformed workflow IDs; no config or API changes.

<sup>Written for commit c57102245c21b21cf84a526fd40d754d016407ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/workflows.test.ts
+++ b/packages/farm/src/__tests__/workflows.test.ts
@@ -238,6 +238,32 @@ describe("Farm workflows", () => {
     expect(handlerSource).toContain('getHeader(event, "x-farm-workflow-secret")');
     expect(handlerSource).toContain("authorization.match(/^Bearer");
     expect(handlerSource).not.toContain('searchParams.get("secret")');
+    expect(handlerSource).toContain("function decodeRouteSegment(segment)");
+    expect(handlerSource).not.toContain("decodeURIComponent(event.context.params");
+  });
+
+  it("404s on a malformed percent-encoded workflow id instead of throwing", async () => {
+    const workflow = {
+      id: "sync-users",
+      filePath: "/virtual/sync-users.ts",
+      description: "Sync users.",
+      schedule: [],
+      routePath: "/api/_farm/workflows/sync-users",
+    };
+    const handler = createFarmWorkflowRequestHandler({
+      workflows: [workflow],
+      config: resolveWorkflowsConfig({ secret: "test-secret" }),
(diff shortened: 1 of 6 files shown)


kira-1011:
Folded into #515 to keep it one review.

Issue #492, opened 2026-08-24
loadSearchParams drops every repeated value but the first, so asArrayOf never sees them


surafel58:
## Summary

`loadSearchParams` appends every value of a repeated parameter and then reads back only the
first, so the multi-value branch it just built can never take effect.

`packages/farm/src/query/server.ts`, the two halves:

```ts
// :55-63  collects every value
if (Array.isArray(value)) {
  for (const v of value) {
    urlParams.append(key, v);
  }
}
```

```ts
// :69  reads one value
const value = urlParams.get(key);
```

`URLSearchParams.get` returns the first match. `getAll` returns all.

## Reproduction

```ts
import { describe, it } from "vitest";
import { asArrayOf, asString } from "../query/parsers";
import { loadSearchParams } from "../query/server";

describe("loadSearchParams", () => {
  it("keeps every value of a repeated parameter", async () => {
    console.log("array prop      ->", JSON.stringify(
      await loadSearchParams(Promise.resolve({ tag: ["react", "vite", "zod"] }), {
        tag: asArrayOf(asString),
      }),
    ));
    console.log("URLSearchParams ->", JSON.stringify(
      await loadSearchParams(Promise.resolve(new URLSearchParams("tag=react&tag=vite&tag=zod")), {
        tag: asArrayOf(asString),
      }),
    ));
    console.log("comma string    ->", JSON.stringify(
      await loadSearchParams(Promise.resolve({ tag: "react,vite,zod" }), {
        tag: asArrayOf(asString),
      }),
    ));
  });
});
```

Output:

```
array prop      -> {"tag":["react"]}
URLSearchParams -> {"tag":["react"]}
comma string    -> {"tag":["react","vite","zod"]}
```

`vite` and `zod` are dropped in the first two. Only the comma form works, so `asArrayOf` is
reachable only through a shape the framework does not produce.

## Why the broken shapes are the normal ones

`loadSearchParams` accepts `Record<string, string | string[] | undefined>`, which is the
`searchParams` prop a page receives. Since #365 that prop collects repeated keys into arrays at
every site, so the array form is what a page actually hands to this function. The
`URLSearchParams` overload has the same problem.

The docs describe `asArrayOf(parser)` as "Repeated values serialized through another parser"
(`docs/src/app/docs/query/page.md:111`), and `asArrayOf.serialize` joins with `,`
(`query/parsers.ts:84`), so serialize and parse do not round-trip for repeated input either.

## Suggested fix

[...]

PR #525, opened 2026-08-25
fix(core): keep repeated query parameters in workflow GET triggers

### LINKED ISSUE #524: Workflow GET triggers drop repeated query parameters
## Summary

A GET workflow trigger builds its payload with `Object.fromEntries(searchParams.entries())`,
which keeps only the last value of a repeated query parameter. The earlier values are dropped
silently.

```ts
// packages/farm/src/workflows.ts:705, the dev and node handler
return Object.fromEntries(url.searchParams.entries());

// packages/farm/src/workflows.ts:641, the emitted Nitro handler
return Object.fromEntries(event.url.searchParams.entries());
```

Core already has one representation for this. `searchParamsToObject` in
`packages/farm/src/search-params.ts` collects repeated keys into arrays, and its own docstring
says the dev renderer, the production SSR entry, the SPA page-data endpoint and the generated
client hydration runtime all share it "so every environment agrees on one representation".
`searchparams.test.ts` asserts that across environments. The workflow trigger is not one of
them, so a scheduler or a cron caller passing a repeated parameter loses data that every other
entry point preserves.

Same class as #492 and #498.

## Reproduction

An app with `src/workflows/hello.ts` echoing its payload:

```ts
export default {
  id: "hello",
  async run(ctx) {
    return { payload: ctx.payload };
  },
};
```

`farm dev`:

```
GET /api/_farm/workflows/hello?tag=a&tag=b
  {"id":"hello","ok":true,"result":{"payload":{"tag":"b"}}}

GET /api/_farm/workflows/hello?tag=a
  {"id":"hello","ok":true,"result":{"payload":{"tag":"a"}}}
```

Expected for the first, matching every other entry point:

```json
{"tag": ["a", "b"]}
```

The single-value case is the control, it shows the payload path is otherwise fine.

## Suggested fix

Use the shared helper at both sites. The emitted handler already imports from
`@farm.js/core/internal/production-runtime`, which re-exports `searchParamsToObject`, so it is
a one line addition to the existing import rather than an inlined copy.

## Related, not verified at runtime

`packages/farmjs-plugin/src/rsc/entries/rsc.ts:744` and `:911` build the page `searchParams`
prop with the same `Object.fromEntries(url.searchParams)`. If the RSC entry is a live path that
looks like the same inconsistency, since `searchParams` is exactly the prop #365 normalized and
`loadSearchParams` consumes. I have only read the code there, so I did not want to fold it into
this report. Happy to check it separately.

## Environment

- `@farm.js/core` `0.1.0-beta.61`, `main` at `099e60d0`
- Node.js 24, not platform specific

---

@Kinfe123, I'm working on the PR for this! Could you please assign this issue to me?

kira-1011:
Closes #524.

The GET trigger payload was built with `Object.fromEntries(searchParams.entries())`, which
keeps only the last value of a repeated key, so `?tag=a&tag=b` reached the workflow as
`{ tag: "b" }` with the first value dropped silently.

Both handlers now use `searchParamsToObject`, the representation `search-params.ts` already
documents as shared by the dev renderer, the production SSR entry, the SPA page-data endpoint
and the generated client runtime. The emitted Nitro handler pulls it from
`@farm.js/core/internal/production-runtime`, which it was already importing from, so no inlined
copy was needed.

Same class as #492 and #498.

## Verification

`farm dev` against an app whose workflow echoes its payload:

```
                            before                  after
?tag=a&tag=b                {"tag":"b"}             {"tag":["a","b"]}
?tag=a                      {"tag":"a"}             {"tag":"a"}
?tag=a&tag=b&one=x          {"tag":"b","one":"x"}   {"tag":["a","b"],"one":"x"}
no params                   {}                      {}
```

The single value and empty cases are unchanged, only the repeated case differs.

Two tests, both failing on `main` first: a behavioural one on the request handler, and an
emitted-source assertion added to `searchparams.test.ts` alongside the existing per-environment
checks, so the workflow trigger is now covered by the same invariant as everything else.

Core suite green, 1257 passed and 6 skipped, no failures. Type check and lint clean.

## Note

`packages/farmjs-plugin/src/rsc/entries/rsc.ts:744` and `:911` build the page `searchParams`
prop the same way. I flagged it in #524 but left it out here, since I have only read that code
and not exercised the RSC path. Happy to look at it separately.

`middleware/production-runtime.ts:841` looked like the related #507 shape but is correct, it
uses `getSetCookie` and only falls back to `.get()` on runtimes without it.

### DIFF
--- a/packages/farm/src/__tests__/searchparams.test.ts
+++ b/packages/farm/src/__tests__/searchparams.test.ts
@@ -224,6 +224,19 @@ describe("SearchParams in the production runtime", () => {
     expect(source).not.toContain("const searchParams: Record<string, string> = {};");
   });
 
+  it("triggers workflows with array-aware search params", () => {
+    const source = readSource("workflows.ts");
+
+    expect(source).toContain('import { searchParamsToObject } from "./search-params";');
+    expect(source).toContain("return searchParamsToObject(url.searchParams);");
+    // the emitted Nitro handler pulls it from the same re-export
+    expect(source).toContain(`  searchParamsToObject
+} from "@farm.js/core/internal/production-runtime";`);
+    expect(source).toContain("return searchParamsToObject(event.url.searchParams);");
+    expect(source).not.toContain("Object.fromEntries(url.searchParams.entries())");
+    expect(source).not.toContain("Object.fromEntries(event.url.searchParams.entries())");
+  });
+
   it("shares the dev renderer's helper", () => {
     const source = readSource("server", "renderer.ts");
 
--- a/packages/farm/src/__tests__/workflows.test.ts
+++ b/packages/farm/src/__tests__/workflows.test.ts
@@ -266,6 +266,32 @@ describe("Farm workflows", () => {
     }
   });
 
+  it("keeps every value of a repeated query parameter in a GET trigger payload", async () => {
+    const workflow = {
+      id: "sync-users",
+      filePath: "/virtual/sync-users.ts",
+      description: "Sync users.",
+      schedule: [],
+      routePath: "/api/_farm/workflows/sync-users",
+    };
+    const run = vi.fn(async (ctx: any) => ({ received: ctx.payload }));
+    const handler = createFarmWorkflowRequestHandler({
+      workflows: [workflow],
+      config: resolveWorkflowsConfig({ secret: "test-secret" }),
+      loadModule: async () => ({ default: { run } }),
+    });
+
+    const response = await handler(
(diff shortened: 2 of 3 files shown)


Kinfe123 (approved):
verified locally, both handlers and the emitted import check out. merging. but we need a follow up on being vurl on `__proto__` which all js object secretly has a parent they inherit from, that is prototype and it has a magical property name `__proto__` , so any attacker can do and swap their stuff via this behavior so i will up witha that fix + some rsc code lacks the PR fix behavior

PR #545, opened 2026-08-26
feat(sentry): add a Sentry plugin for framework lifecycle observability

### LINKED ISSUE #544: Sentry plugin for framework lifecycle observability
## Summary

Add Sentry support as a published plugin. Following the direction in the thread, this is a
plugin rather than an integration, since Sentry only observes framework lifecycles and does not
own routes or application APIs.

Two entry points:

- `sentryPlugin()` for Farm event mapping and build behavior
- `registerSentry()` for early runtime initialization

## sentryPlugin()

```ts
import { defineConfig } from "@farm.js/core";
import { sentryPlugin } from "@farm.js/plugin-sentry";

export default defineConfig({
  plugins: [
    sentryPlugin({
      dsn: process.env.SENTRY_DSN,
      environment: process.env.NODE_ENV,
      tracesSampleRate: 0.1,
      sendDefaultPii: false,
    }),
  ],
});
```

Hook mapping:

| Hook | Behavior |
| --- | --- |
| `runtime.start` | initialize the SDK for the server runtime |
| `runtime.context` | open a per request scope and expose the trace id |
| `runtime.before` / `runtime.after` | open and close the request span |
| `runtime.error` | `captureException` with route and request context |
| `render.before` / `render.html` | render spans |
| `build.configure` | enable source maps on the Nitro config |
| `build.after` | upload source maps and create the release |
| `runtime.close` | flush pending events |

`client` covers browser errors and performance later, once the server side is settled.

Defaulting `sendDefaultPii` to false, since error events can carry request headers and user data.

## registerSentry()

```ts
// src/instrumentation.ts
import { registerSentry } from "@farm.js/plugin-sentry";

export const register = registerSentry({ dsn: process.env.SENTRY_DSN });
```

This exists because the Node SDK has to load before the rest of the app so its auto
instrumentation can patch modules correctly. Plugin `setup` cannot guarantee that ordering.

## Runtime support

Node presets first. `FarmInstrumentationContext` already reports `runtime` as `nodejs`, `edge`
or `bun`, so the plugin can gate on that and no new convention is needed.

Edge comes later as separate work. `@sentry/node` does not run on Cloudflare Workers,
`@sentry/cloudflare` is the supported SDK there, and there is an open report of workerd export
conditions resolving to `@sentry/node` and pulling `node:*` built-ins into the Worker bundle
(getsentry/sentry-javascript#20038).

## Notes

- Serverless hosts do not always give a clean shutdown, so on those presets the plugin flushes
  at the end of the request instead of relying on `runtime.close`.
- SDK init stays out of `setup`, since `setup` can run in a build manager and again in a
  deployed runtime and needs to stay deterministic.
- The `runtime.context` key will be namespaced, since duplicate top level context keys fail the
  request.

## Open

- Package name, `@farm.js/plugin-sentry` based on the `farm-plugin-*` convention in the plugin
  docs. Happy to use something else.
- Whether this should stay separate from `@farm.js/otel` or share anything with it.

kira-1011:
Closes #544.

Adds `@farm.js/sentry`, a framework plugin for Sentry error reporting and request tracing. `sentryPlugin()` maps Farm's runtime and build lifecycles, while `registerSentry()` initializes the Node SDK early from `src/instrumentation.ts` so automatic instrumentation can attach before the application loads.

## Hook mapping

| Hook | Behavior |
| --- | --- |
| `runtime.start` | initializes the SDK safely and subscribes to Farm's unfiltered internal error stream |
| `runtime.context` | reuses the active Sentry request span, or creates a fallback transaction |
| `runtime.after` | records response status and closes plugin-owned spans |
| `runtime.error` | captures failures that escape the request pipeline |
| `runtime.close` | unsubscribes and flushes |
| `build.configure` | enables source maps when requested |

Internally handled failures such as render errors never reach `runtime.error`, so the plugin also observes Farm error events. Deduplication is shared across plugin instances but expires after the current event-loop turn, preventing duplicate captures without permanently suppressing a reused `Error` object. Thrown primitive values use the same short-lived deduplication window. The internal subscription bypasses application event filters without changing the user's observability configuration.

The plugin accepts an injected SDK through `sdk`, or loads the optional `@sentry/node` peer through a literal dynamic import so serverless bundlers include it. Missing or broken monitoring never prevents the application from starting or shutting down. Request, runtime-shutdown, and instrumentation-shutdown flushes share one protected path that contains rejected promises and synchronous SDK failures. PII remains disabled by default, route patterns keep span names low-cardinality, and response flushing is available for short-lived hosts.

## Verification

- 47 Sentry plugin tests
- 19 core observability and development tracing tests
- package type-check and build
- formatting and lint checks

The package is included in `bump.config.ts`. Edge and Bun SDK support remain separate work.

## Documentation

- adds a dedicated `/docs/plugins/sentry` guide covering installation, DSN configuration, early instrumentation, serverless flushing, SDK injection, options, source maps, and runtime support
- organizes `Plugin Ecosystem` into `Official Plugins` (`Sentry`) and `Plugin Authoring` (`Create a Plugin`, `Client Plugin API`) sidebar groups
- adds the Sentry brand icon and updates generated route types


Kinfe123 (changes requested):
Thanks for working on this. I think a plugin is the right place for Sentry, and the overall direction looks good.

A few things should be fixed before merging:

1. `startInactiveSpan` never becomes active, so automatic HTTP and database spans will not be connected to the request trace.
2. `@sentry/node` is not declared as a dependency. If it is missing or broken, the plugin silently does nothing. I think it needs a peer dependency, a real SDK test, and a clear error.
3. The documented setup can call `Sentry.init()` twice through `registerSentry` and `runtime.start`.
4. `flushOnResponse` does not run when the request throws, which can lose exceptions on serverless hosts.
5. `sourceMaps: true` only generates maps. It does not upload them to Sentry, so the README currently promises more than the option provides.

I would also rename `client` to `sdk`, make sure observability filters cannot hide errors from the plugin, and add the package to `bump.config.ts`. Once these are addressed, I think this will be a useful addition.

kira-1011:
Thanks, all five were right. Pushed in 0794f5d8.

**1. Active span.** The plugin now takes the span the SDK already opened, through `getActiveSpan`
and `getRootSpan`, and renames it with `updateSpanName`. It only starts its own span when nothing
is active, and only ends spans it created, so the SDK keeps ownership of its own. Automatic HTTP
and database spans stay nested under the request now.

**2. `@sentry/node`.** Declared as an optional peer dependency. When a `dsn` is set and nothing
resolves, it throws with an actionable message instead of reporting nothing. I also added the SDK
as a dev dependency and a test that asserts every function the plugin calls exists on the real
module, so an upstream rename fails here rather than showing up as missing data.

**3. Double init.** `initSentryOnce` checks `getClient()` first, so `registerSentry` and
`runtime.start` cannot both initialize. It also leaves an application's own `Sentry.init` alone.

**4. `flushOnResponse` on failure.** Now flushed from `runtime.error` too. You were right that a
failing request never reaches `runtime.after`.

**5. Source maps.** The README now says plainly that the option generates maps and does not upload
them, and that uploading from the build is not implemented.

Also done: renamed `client` to `sdk`, added the package to `bump.config.ts`, and the observability
point. That one was a real hole. `shouldEmitFarmEvent` checks the `events` allowlist before
handlers run, so an app narrowing it for logging would have silently stopped errors from reaching
Sentry. The `configure` hook now adds the error types back.

## Verification

34 unit tests, type check and lint clean.

Checked again against a real Sentry project after the span change, since it was the riskiest part.
With `debug` on, the SDK reports:

```
[Tracing] Starting sampled root span
[Tracing] Finishing "http.server" root span "GET /" with ID b8ea7ea4b388694d
[Tracing] Finishing "http.server" root span "GET /technologies" with ID b6eb2f2abffda3f8
```

Route named, sampled, sent. Errors still arrive tagged with the route.

One small thing I noticed while testing: without `release` set the SDK logs `Discarded session
because of missing or non-string release`. Harmless, and the option already exists, so I left it.

Kinfe123:
have you run it with working sentry dns and credentials ?

kira-1011:
Yes, against a real Sentry project with a working DSN, in both dev and a production build. Ran
it again just now on a real app after the span change, with the SDK's own `debug` output as
evidence rather than going by the dashboard.

Routes exercised: two static pages, a dynamic route hit with two different ids, a page that
throws during render, an API route that throws, and an existing API route that succeeds.

## Traces

Every request produced a transaction, in both dev and production:

```
2 root span "GET /"
1 root span "GET /technologies"
2 root span "GET /sentry-test/[id]"
1 root span "GET /sentry-test/throw"
1 root span "GET /api/sentry-test"
```

`/sentry-test/abc` and `/sentry-test/xyz` collapse to one name, so the route pattern naming is
doing its job.

## Errors

Production captured both:

```
Captured error event `sentry-test: page render failure`
Captured error event `sentry-test: api route failure`
```

Exactly two captures for two errors, so the deduplication holds now that the reported set is
shared across plugin instances rather than per setup state. That was a real bug, thanks for
prompting the deeper check.

## One difference worth flagging

Dev captured the page render failure but not the API route failure. Production captured both.

`api.error` is only emitted from `nitro/universal-build.ts`, so in dev a throwing API route logs
`[API Error]` to the console and no event is emitted. Nothing for the plugin to observe. Traces
are unaffected, the transaction is still recorded either way.

That looks like a gap in Farm rather than something the plugin can work around, since no plugin
hook sees it either. Happy to open a separate issue for it if you agree, and it does not block
this PR.

## Also in this branch

`7aa20fb9` moves the reported error set to module scope. Sentry's client is process wide, so two
plugin instances, or a `setup` that runs more than once in a process, were reporting the same
error twice. There is a test for it.

35 tests, type check and lint clean.

Still unverified: `flushOnResponse` on a host that kills the process mid request. It fires on
both the response and error paths locally, but only a real serverless deploy exercises the case
it exists for.

Kinfe123:
nice catch. I confirmed this is a real dev-runtime gap. The Vite path does have an `api.error` handler, but `APIRouteManager.getHandler()` catches the exception first and converts it into a 500 response, so the outer observability and plugin hooks never receive it. Production uses a different handler and reports it correctly.

I agree this should be tracked and fixed separately with a development API regression test, and it does not need to block this PR. Please open an issue for it.

kira-1011:
@Kinfe123 this is ready for another look. Three more fixes since your commit, each reproduced
before being changed.

**Never take the app down.** `assertSentrySdk` threw when the SDK was missing. That runs during
instrumentation, so on Vercel the function never booted and every route returned 500. It now
logs and continues.

**Literal import specifier.** The dynamic import used a variable, which bundlers cannot trace,
so `@sentry/node` was never included in the serverless output. The plugin reported a missing
SDK even though the package was installed. This was the actual reason nothing reached Sentry.

**Contain flush failures.** A rejected flush was handed to `waitUntil` uncaught, and Farm passes
that straight to the host when one supplies it, which is the serverless case. Reproduced against
`PluginManager`, where it fires a real `unhandledRejection`. `runtime.close` had the same problem
and failed shutdown. Both now log and continue.

**Deduplicate thrown values that are not Errors.** A `WeakSet` cannot hold a string, so a thrown
string reaching both the event stream and `runtime.error` was reported twice while an Error was
reported once.

## Verified on a real deploy

Vercel preview with a working DSN, not only unit tests:

- a page and an API route that throw both arrive in Sentry, tagged `farm.event` and `farm.route`,
  with `handled: yes` so it is the plugin capturing and not a global handler
- transactions are named by route pattern, `GET /users/[id]` rather than the pathname
- a thrown string produced 3 events for 3 requests, where before the fix it produced 6
- the app stayed up through six deliberate failures with `flushOnResponse` on

46 tests, all new ones checked to fail without their fix. CI green.

## Two limits worth knowing

The literal specifier test checks the source text. It catches a revert, not a different bundling
regression. Only a deploy does that.

The flush failure path is covered by unit tests through `PluginManager`, including a real
`unhandledRejection` listener, but the successful flushes on the deploy did not exercise the
catch itself.

Kinfe123 (approved):

PR #558, opened 2026-08-28
fix(tests): stop the Windows production middleware test hanging in cleanup

### LINKED ISSUE #434: Windows: production middleware test hangs in fixture cleanup

kira-1011:
Closes #434.

## What it was

Not the wasm image codec, despite how the issue was originally titled. Every request in the
test finishes quickly, including the one that loads the codec, at 3ms, and the whole test body
runs in under 9 seconds. The 120s went entirely to fixture teardown.

Bisecting the tree found one file:

```
EPERM: operation not permitted, unlink
  .vercel/output/functions/__nitro.func/node_modules/@img/sharp-win32-x64/lib/sharp-win32-x64.node
```

The build copies sharp's native binary into the output and the test process loads it while
optimizing images. Windows will not unlink a loaded module, so the recursive remove retries
that file at every parent directory on the way up and never finishes.

## The change

On Windows, remove what is removable and leave the locked file. Each run also sweeps fixtures
an earlier run could not finish, which works because the lock is released when that process
exits. One 428KB file survives a run and the next run clears it. Other platforms keep the
existing recursive remove.

The Windows skip on the test is no longer needed.

## Verification

The test passes on Windows in about 8s, previously a 120s timeout. Core suite is green at 1268
passed and 0 failed, with skipped going from 6 to 5 because this test now runs.

## Ruled out

Two explanations that looked right and were not:

- `fs.rm` following the `node_modules/@farm.js/core` junction into `packages/farm`. It does not
  follow junctions, the target is untouched and it returns in 0ms.
- The dynamically imported entry holding a lock. Importing an `.mjs` and deleting it from the
  same process takes 2ms.

### DIFF
--- a/packages/farm/src/__tests__/fixtures/middleware-production-fixture.ts
+++ b/packages/farm/src/__tests__/fixtures/middleware-production-fixture.ts
@@ -9,7 +9,7 @@ const packageRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "
 const FIXTURE_PREFIX = ".tmp-production-middleware-";
 
 export async function createMiddlewareProductionFixture(): Promise<string> {
-  await removeStaleFixtures();
+  if (process.platform === "win32") await removeStaleFixtures();
   const root = await fs.mkdtemp(path.join(packageRoot, FIXTURE_PREFIX));
 
   await fs.mkdir(path.join(root, "node_modules", "@farm.js"), {
@@ -462,21 +462,23 @@ export async function cleanupMiddlewareProductionFixture(root: string): Promise<
   // module, so that one file is undeletable until the process exits. A
   // recursive remove retries it at every level on the way up and never
   // finishes, which used to exhaust the whole test timeout. Remove what is not
-  // locked instead and leave the rest for the operating system.
+  // locked instead. Exiting releases the lock, and the next run sweeps what is
+  // left behind.
   await removeUnlockedEntries(root);
 }
 
 /**
  * Sweep fixtures an earlier run could not finish removing.
  *
- * On Windows the locked native binary is released when that process exits, so
- * the leftovers are deletable by the time the next run starts.
+ * Windows only. The process that held the native binary has exited by now, so
+ * its leftovers are removable.
  */
 async function removeStaleFixtures(): Promise<void> {
   let entries: Dirent[];
   try {
     entries = await fs.readdir(packageRoot, { withFileTypes: true });
-  } catch {
+  } catch (error) {
+    if (!isLockError(error)) throw error;
     return;
   }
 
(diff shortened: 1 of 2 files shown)


Kinfe123 (commented):
Thanks for pinning this down. i think the diagnosis and Windows test result make the approach valid.

Two small hardening suggestions:

- Run `removeStaleFixtures()` only on Windows. Other platforms already remove the fixture completely, and scanning every prefixed directory there is unnecessary.
- Ignore only expected Windows lock errors such as `EPERM`, `EBUSY`, and `ENOTEMPTY`; rethrow unexpected errors so cleanup regressions are not silently hidden.

One wording nit: the OS releases the lock when the process exits, but the next run-not the OS-emoves the leftover file. A child process would allow complete same-run cleanup, though I do not think that larger change is required for this test-only fix.

kira-1011:
Thanks, all three applied in 859c5ad8.

- `removeStaleFixtures()` now runs only on Windows.
- Lock errors are matched explicitly, `EPERM`, `EBUSY`, `ENOTEMPTY` and `ENOENT`, and anything
  else is rethrown. That covers the `readdir` guards too, which were swallowing everything.
- Comment corrected. Exiting releases the lock, the next run removes the file.

On the child process point, I agree it is the more complete fix and out of scope here. Worth
recording why the smaller fix is not just a workaround: the lock is held for the lifetime of the
process because Node cannot unload a native addon, so retrying cannot help. I also checked
rimraf's rename trick and it does not release the lock either:

```
unlink at original name: FAILED EPERM
renamed ok
unlink AFTER rename:     FAILED EPERM
```

Renaming moves the name, not the handle. That is why the retry and move strategies that work for
transient Windows locks, antivirus or the indexer, do not apply to this one.

Test passes in about 7.5s. Core suite green at 1268 passed and 0 failed.

Kinfe123 (approved):

PR #564, opened 2026-08-28
fix(dev): report a layout that fails to load instead of dropping it quietly

### LINKED ISSUE #560: A layout module that fails to load is dropped silently, deleting server-rendered markup
## Summary

When a layout module fails to evaluate in the browser, `loadLayoutComponents` catches the
error, logs a warning, and drops the layout from the client tree. React then finds the server
HTML does not match, regenerates the branch, and the layout markup is removed from the DOM. A
visible header disappears the moment hydration runs.

`packages/farm/src/vite.ts:3970-3990`, inside the emitted client runtime:

```js
async function loadLayoutComponents(layouts = []) {
  const loadedLayouts = [];

  for (const layout of layouts) {
    if (!layout?.modulePath) continue;
    try {
      const layoutModule = await import(/* @vite-ignore */ layout.modulePath);
      ...
      if (LayoutComponent) loadedLayouts.push({ ...layout, Component: LayoutComponent });
    } catch (error) {
      console.warn('[Farm.js] Could not load layout:', layout.modulePath, error);
    }
  }

  return loadedLayouts;
}
```

The layout is omitted from `loadedLayouts`, so `wrapWithLoadedLayouts`
(`packages/farm/src/vite.ts:3992-4007`) produces a tree with no layout boundary.

Any error in a layout module produces this, not just the one below. A typo or a bad import has
the same effect: server rendered markup is deleted and the only signal is a warning.

## Reproduction

One way to make a layout module throw in the browser is to read `process.env` at module scope,
directly or through an import.

`src/app/layout.tsx`:

```tsx
import type { LayoutProps, Metadata } from "@farm.js/core";
import { HeaderNav } from "../components/header-nav.tsx";

export const metadata: Metadata = {
  title: "repro",
  metadataBase: new URL(process.env.SITE_URL ?? "http://localhost:3000"),
};

export default function RootLayout({ children }: LayoutProps) {
  return (
    <div>
      <header id="site-header">
        <HeaderNav />
      </header>
      {children}
    </div>
  );
}
```

`src/components/header-nav.tsx`:

```tsx
"use client";
import { usePathname } from "@farm.js/core/navigation";

export function HeaderNav() {
  return <nav>HEADER path={usePathname()}</nav>;
}
```

The page must also be a client page, so the served HTML carries both `data-farm-client="true"`
and `data-farm-layout-client="true"` on `#__farm_page__`. A server only layout is never
re-imported on the client and does not show this.

Run `farm dev` and open `/`.

## Expected

A layout that fails to load is reported clearly, and the failure does not silently delete
markup that was rendered on the server.

## Actual

The server HTML contains the header. After hydration the DOM has zero `<header>` elements and
zero `[data-farm-layout-boundary="true"]` elements. The only signal is:

```
[Farm.js] Could not load layout: /src/app/layout.tsx ReferenceError: process is not defined
Hydration failed because the server rendered HTML didn't match the client.
```

## Note on process.env, corrected

An earlier version of this issue suggested Farm should apply the `process.env` replacement in
dev that the production client build applies. That is not right, and I want to correct it
before anyone spends time there.

Vite skips define replacement for client modules in dev by design,
`vite:define` in `vite/dist/node/chunks/dep-*.js`:

```js
async transform(code, id, options) {
  const ssr = options?.ssr === true;
  if (!ssr && !isBuild) {
    return;
  }
```

So `process.env` is unavailable in browser code during dev in any Vite app, and adding a define
in Farm would not change it, because the plugin returns before user defines are considered.
Reading `process.env` in a client reachable module is an application level mistake and
`import.meta.env` is the supported way.

That leaves the part above as the Farm issue: the failure mode. A layout module that throws
should not disappear along with the markup it produced.

## Environment

- `@farm.js/core` and `@farm.js/cli` 0.1.0-beta.69
- Node.js 24.15.0, Windows 11

@Kinfe123, happy to work on a PR for this if you would like it assigned.

kira-1011:
Closes #560.

## Problem

A layout module that throws while being imported on the client is caught, warned about, and
skipped.

Skipping it leaves the client tree different from the server's, so React discards the server
rendered markup for that branch. Anything the layout drew, a header for example, disappears the
moment hydration runs. The only signal is a `console.warn`.

Any error in a layout module causes this, not just one kind. A typo or a bad import has the same
effect.

## Change

The layout and loading boundary failures now report through `console.error` and state the
consequence:

```
[Farm.js] Layout failed to load and was skipped: /src/app/layout.tsx. The client tree no
longer matches the server, so React will discard the server rendered markup for this route,
including anything this layout renders.
```

This does not make a broken layout work. It replaces a silent loss of UI with a message that
names the cause.

## Note on the process.env part of the issue

The issue first suggested Farm should apply the `process.env` replacement in dev that the
production client build applies. That is not correct, so it is not part of this change.

Vite skips define replacement for client modules in dev by design:

```js
async transform(code, id, options) {
  const ssr = options?.ssr === true;
  if (!ssr && !isBuild) {
    return;
  }
```

Adding a define in Farm would not help, because the plugin returns before user defines are
considered. Reading `process.env` from a client reachable module is an application level
mistake, and `import.meta.env` is the supported alternative. The issue has been corrected.

## Tests

Two, both failing without this change:

1. The layout and loading boundary failures report through `console.error` and name the
   consequence.
2. The message contains no newline escape. The surrounding template literal would consume it and
   emit a real newline inside a JS string in the generated client runtime.

## Verification

Reproduced on `farm dev` with a layout whose module throws in the browser. Before, the only clue
was a warning and a header that vanished. After correcting the module, the layout survives
hydration: `header` and `[data-farm-layout-boundary="true"]` both go from 0 to 1.

Core suite green, 1270 passed and 0 failed. Type check and lint clean.

### DIFF
--- a/packages/farm/src/__tests__/client-component.test.ts
+++ b/packages/farm/src/__tests__/client-component.test.ts
@@ -2,7 +2,7 @@ import { pathToFileURL } from "node:url";
 import fs from "fs";
 import os from "os";
 import path from "path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import {
   getClientModuleMetadata,
   getIslandStrategyExport,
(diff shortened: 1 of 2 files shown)


Kinfe123 (approved):
I verified the diagnostic path, fixed the loading-boundary error reporter so it preserves the original failure, and added an executable regression test for the emitted hydration helper.

Issue #565, opened 2026-08-28
Support isolated client hydration inside server-rendered layouts


Kinfe123:
## Summary

In the standard non-RSC React renderer, a layout that contains interactive client code can become the hydration boundary. Farm then imports and executes the complete layout module in the browser so React can recreate the tree.

This is valid traditional SSR behavior, but it also means server-oriented layout code enters the browser graph. Issue #560 exposed one consequence: module-scope `process.env` access can break the layout during development hydration.

## Goal

Allow a server-rendered layout to contain explicit client components without requiring the complete layout module to execute in the browser.

```text
Server-rendered Layout
├── static layout markup
├── hydrated Client Component
└── server-rendered page
```

This should apply to the standard React renderer without requiring RSC.

## Expected behavior

- Importing a `"use client"` child does not automatically place its server layout in the browser graph.
- Farm emits an isolated hydration marker, module reference, and serialized props for the client boundary.
- Only the explicit client boundary is hydrated when the composition is supported.
- Nested islands, SPA navigation, Suspense, errors, and context behavior are defined and tested.
- Unsupported compositions either fall back safely to full-tree hydration or produce a clear diagnostic.

## Notes

This is intentionally separate from #560. That issue should remain focused on development/production environment parity and surfacing layout import failures.

Farm already tracks island hydration strategies, so the existing island runtime may provide part of the implementation. This should begin as a focused design/RFC before changing default hydration behavior.

Kinfe123:
Follow-up: #564 has now landed, so layout/loading import failures are reported clearly and covered by an executable regression test. This issue remains the separate architectural step; there is no implementation PR linked yet. Before changing the default hydration path, the implementation should include a fixture proving the server layout stays out of the browser graph, plus coverage for nested islands, navigation, context, Suspense/errors, and the documented fallback for unsupported compositions.

Kinfe123:
RFC follow-up opened in #568. It defines the server proxy and boundary marker protocol, serialization and fallback rules, isolated-root navigation/cleanup semantics, the experimental rollout flag, and the dev/production acceptance matrix. This intentionally keeps #565 open for the phased runtime implementation after the contract is reviewed.

Kinfe123:
Update: the first opt-in implementation landed in #568 and is available behind `experimental.isolatedClientHydration`. I split the remaining RFC acceptance work into focused follow-ups:

- #735: SPA navigation, retained state, root disposal, and stale navigation work
[...]

PR #568, opened 2026-08-28
feat(react): add experimental isolated client hydration

Kinfe123:
## Summary

- add `experimental.isolatedClientHydration` with `"off"`, `"analyze"`, and `"enabled"` modes for non-RSC React applications
- keep `"use client"` valid without RSC while moving eligible local leaves into independently hydrated React roots
- preserve route-wide hydration for explicit client owners, package boundaries, ambiguous exports, and mixed ownership graphs
- serialize only safe props, preserve SSR output when serialization or hydration fails, and dispose isolated roots during navigation
- keep the default bundle free of isolated markers/runtime and make analyze output byte-identical to off
- document the configuration, safety rules, rollout contract, and performance gates

## Performance validation

The production fixture builds the same application in enabled and off modes. It verifies that enabled mode excludes a large server-layout sentinel from the browser graph and produces strictly less total client JavaScript in both raw bytes and gzip bytes. The server still renders the complete layout and the client leaf remains interactive.

## Validation

- core TypeScript check
- core test suite: 159 files passed, 1,282 tests passed, 1 skipped
- focused production build/SSR comparison for enabled, analyze, and off
- real browser SSR and click test for an isolated `"use client"` counter
- `oxlint` on changed TypeScript: 0 errors; one existing unrelated unused-variable warning
- `git diff --check`

Refs #565


Kinfe123:
Added explicit RFC-first rationale, fail-closed safety invariants, resource bounds, and measurable performance gates in cd721af3. The proposal now requires a strict client-JS reduction for the representative layout fixture, CPU/interaction non-regression, SSR/HTML/request/heap reporting, and a many-boundary crossover benchmark. If isolation does not demonstrate a net gain for a route shape, the existing route-wide hydration plan remains the fallback.

PR #603, opened 2026-08-30
fix(cache): serialize invalid dates to a stable marker instead of throwing

### LINKED ISSUE #576: Invalid Date in a cache key throws RangeError before the cached function or server query runs
## What happens

An invalid `Date` anywhere in a cache key makes `createFarmCacheKey` throw `RangeError: Invalid time value`. Because key building runs before anything else, `unstable_cache` throws before the wrapped function runs and `useServerQuery` / `fetchServerQuery` throw while computing the call key, so the component render fails with an error from inside `@farm.js/core` instead of the app's own input validation.

An invalid `Date` is easy to get from URL input: `new Date(searchParams.get("from"))` returns an Invalid Date for any malformed value, it does not throw.

## Repro

```ts
import { createFarmCacheKey, unstable_cache } from "@farm.js/core";
import { createServerQueryCallKey } from "@farm.js/core/server-query-runtime";

createFarmCacheKey([new Date("not-a-date")]);
// RangeError: Invalid time value

let calls = 0;
const report = unstable_cache(async (from: Date) => { calls++; return "rows"; }, ["report"]);
await report(new Date("2024-13-45"));
// throws before the function runs, calls === 0

createServerQueryCallKey(query, { from: new Date("2024-02-31x"), page: 1 });
// RangeError: Invalid time value
```

Output on `main` (`0.1.0-beta.69`):

```
invalid Date: THROWS RangeError Invalid time value
unstable_cache(invalid Date): THROWS before fn runs, calls = 0 | Invalid time value
server query key: THROWS RangeError Invalid time value
valid Date key: ["server-query-call",number:1,{"from":date:2024-02-01T00:00:00.000Z,"page":number:1}]
```

## Why

`stableSerialize` in `packages/farm/src/cache.ts` serializes dates with `value.toISOString()`, which throws on an Invalid Date.

## Expected

Key building should never throw on a value it already supports. An invalid date should serialize to a stable marker such as `date:invalid`, so all invalid dates share one key and the caller's own validation decides what to do with the input.

## Related, lower priority

`ArrayBuffer` and `DataView` fall through to the generic object branch and serialize as `{}`, so `new ArrayBuffer(4)` and `new ArrayBuffer(8)` share a key (same class of problem as #569). Typed arrays serialize by index, so `Uint8Array([1])`, `Int8Array([1])` and `{ 0: 1 }` also share a key. Binary values are an unusual cache key input since all framework call sites are fed JSON shaped data, so this may not be worth handling, but it could be done in the same change with a `bytes:` prefix if you want it.

surafel58:
Fixes #576

## Problem

`stableSerialize` calls `value.toISOString()` on every `Date`, and that throws `RangeError: Invalid time value` on an Invalid Date. Key building runs before anything else, so `unstable_cache` threw before the wrapped function ran and `createServerQueryCallKey` threw while computing the call key. An Invalid Date is easy to get from URL input, since `new Date(searchParams.get("from"))` returns one for any malformed value instead of throwing.

## Fix

The `Date` branch now checks `Number.isNaN(value.getTime())` and serializes an invalid date as the stable marker `date:invalid`. All invalid dates share one key, valid date serialization is unchanged, and the caller's own validation decides what to do with the input.

## Tests

New `cache-key-invalid-date.test.ts` covers:

- `createFarmCacheKey` with invalid dates, top level and nested in objects, arrays, sets and maps
- valid date serialization unchanged
- `unstable_cache` runs the wrapped function for an invalid date argument and caches per marker
- `createServerQueryCallKey` builds a key for an invalid date input, equal across invalid dates and distinct from a valid one

4 of 5 fail on `main` with only `cache.ts` reverted (the valid-date test passes either way). Existing `cache.test.ts` and `cache-key-set-map.test.ts` still pass. `tsc --noEmit`, oxlint and oxfmt are clean.

Binary value handling (`ArrayBuffer`, `DataView`, typed arrays) is left out as requested, so this fix stays focused.

---
## Summary by cubic
Fixes cache key building so an invalid `Date` (for example from URL input via `new Date(searchParams.get("from"))`) serializes to a stable `date:invalid` marker instead of throwing `RangeError: Invalid time value`. Valid date serialization is unchanged, and the caller's own validation decides what to do with the invalid input.

**Tests**
- New tests cover invalid dates at the top level and nested in objects, arrays, sets, and maps, plus `unstable_cache` and server query call key behavior.

<sup>Written for commit 2036585d27e22bec7b7ccb5124f94aa99e8062d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/cache-key-invalid-date.test.ts
+++ b/packages/farm/src/__tests__/cache-key-invalid-date.test.ts
@@ -1,0 +1,64 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { configureFarmCache, createFarmCacheKey, unstable_cache } from "../cache";
+import { createServerQueryCallKey } from "../server-query-runtime";
+
+afterEach(() => {
+  configureFarmCache(undefined);
+});
+
+describe("createFarmCacheKey with invalid Date", () => {
+  it("serializes an invalid date to a stable marker instead of throwing", () => {
+    expect(createFarmCacheKey([new Date("not-a-date")])).toBe("[date:invalid]");
+    expect(createFarmCacheKey([new Date(NaN)])).toBe("[date:invalid]");
+    expect(createFarmCacheKey([new Date("2024-13-45")])).toBe(
+      createFarmCacheKey([new Date("also-invalid")]),
+    );
+  });
+
+  it("keeps valid date serialization unchanged", () => {
+    expect(createFarmCacheKey([new Date("2024-02-01T00:00:00.000Z")])).toBe(
+      "[date:2024-02-01T00:00:00.000Z]",
+    );
+  });
+
+  it("serializes an invalid date nested in objects, arrays, sets and maps", () => {
+    const invalid = new Date(NaN);
+    expect(createFarmCacheKey([{ from: invalid, list: [invalid] }])).toBe(
+      '[{"from":date:invalid,"list":[date:invalid]}]',
+    );
+    expect(createFarmCacheKey([new Set([invalid])])).toBe("[set:[date:invalid]]");
+    expect(createFarmCacheKey([new Map([["from", invalid]])])).toBe(
+      '[map:[["from",date:invalid]]]',
+    );
+  });
+
+  it("lets unstable_cache run the wrapped function for an invalid date argument", async () => {
+    let calls = 0;
+    const report = unstable_cache(
+      async (from: Date) => {
+        calls++;
+        return Number.isNaN(from.getTime()) ? "invalid input" : `rows since ${from.toISOString()}`;
+      },
+      ["report"],
+    );
+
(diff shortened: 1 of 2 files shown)


surafel58:
@Kinfe123 this is ready for review. It covers the date:invalid marker, unchanged valid date output, and regression tests for createFarmCacheKey, unstable_cache, and server query call keys. Binary value handling is left out as you asked.

Kinfe123 (approved):

PR #660, opened 2026-09-01
fix(compression): vary identity responses by encoding

Kinfe123:
## Problem

When compression negotiation selected the identity representation, the production compression plugin returned the response without `Vary: Accept-Encoding`. A shared cache could then reuse that uncompressed representation for a client that supports compression. Eligible `HEAD` responses had the same mismatch and could replace the cached metadata for their corresponding `GET` representation during revalidation.

## Root cause

The plugin appended `Accept-Encoding` only inside the branch that produced gzip or Brotli output. It also exited before applying representation metadata to `HEAD` responses.

## Change

Return an equivalent identity response with `Accept-Encoding` appended to its `Vary` header whenever the response is otherwise eligible for compression but no supported encoding is selected. Apply the same variation to eligible `HEAD` responses without attempting to compress them, and preserve a wildcard `Vary: *` value unchanged.

## Safety and compatibility

The response body, status, status text, and existing headers are preserved. Existing `Vary` values are retained without duplicates, including the wildcard form. Responses excluded from compression, including event streams, pre-encoded bodies, range responses, and `no-transform` responses, remain unchanged. There is no public API change.

## Verification

- `pnpm exec oxfmt packages/farm/src/plugins/compression.ts packages/farm/src/__tests__/compression.test.ts`
- `pnpm exec oxlint packages/farm/src/plugins/compression.ts packages/farm/src/__tests__/compression.test.ts`
- `pnpm --filter @farm.js/core exec vitest run src/__tests__/compression.test.ts --reporter=dot` — 6 tests passed
- `git diff --check`

The identity regression test fails before the fix because the response retains only its existing `Accept-Language` variation. The added tests also cover wildcard `Vary` semantics and `HEAD` revalidation metadata.

## Documentation and release

None. This restores HTTP cache semantics without changing configuration or public APIs.

### DIFF
--- a/packages/farm/src/__tests__/compression.test.ts
+++ b/packages/farm/src/__tests__/compression.test.ts
@@ -50,6 +50,23 @@ describe("compression plugin", () => {
     expect(brotliDecompressSync(Buffer.from(await response.arrayBuffer())).toString()).toBe(body);
   });
 
+  it("varies eligible identity responses by Accept-Encoding", async () => {
+    const manager = createManager();
+    const response = await manager.runRuntimeRequest(
+      new Request("https://farm.test/", {
+        headers: { "accept-encoding": "br;q=0, gzip;q=0" },
+      }),
+      () =>
+        new Response("identity", {
+          headers: { vary: "Accept-Language" },
+        }),
+    );
+
+    expect(response.headers.get("content-encoding")).toBeNull();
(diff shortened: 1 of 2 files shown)


Kinfe123:
Addressed the review finding in e1e69bd3: eligible HEAD responses now retain the same `Vary: Accept-Encoding` metadata as their GET representation without being compressed. Added focused HEAD coverage; the compression suite now passes all 6 tests.

PR #701, opened 2026-09-03
fix(navigation): ignore stale generated-router responses

Kinfe123:
## Problem

In a production build, two overlapping navigations can finish out of order. The older generated-router request remains active and can overwrite the DOM, history, scroll position, and navigation state after the newer route has already rendered.

## Root cause

The shared `SPARouter` has an abortable navigation sequence, but the two routers emitted by `universal-build.ts` only expose loading state. They neither abort superseded fetches nor verify ownership after asynchronous plugin, fetch, route-loading, transition, and rendering boundaries.

## Change

Add the shared navigation ownership model to the generated runtime: each navigation gets an incrementing ID and `AbortController`; starting a newer navigation aborts the previous request; every asynchronous commit boundary confirms the navigation is still current. Same-document and document-owned navigation also cancel pending work.

## Safety and compatibility

- Covers both HTML-swap and hydratable production routers.
- Aborts only work superseded by a newer browser intent.
- Preserves plugin lifecycle ordering, view transitions, route interception, history, and scroll behavior for the active navigation.
- Adds no public API and no cost outside the generated client router.

## Verification

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/universal-build-router-runtime.test.ts src/__tests__/client-component.test.ts src/__tests__/spa-router-concurrency.test.ts` (40 tests passed)
- `pnpm exec oxlint packages/farm/src/nitro/universal-build.ts packages/farm/src/__tests__/client-component.test.ts packages/farm/src/__tests__/universal-build-router-runtime.test.ts` (0 errors; one pre-existing unused-variable warning in `universal-build.ts`)
- `git diff --check`

The new generated-runtime regression test fails without the navigation ownership and guarded completion behavior.

## Documentation and release

None. This aligns production-generated routing with the existing shared SPA router contract.

---
## Summary by cubic
Prevents stale navigation responses from overwriting newer routes in production builds by making each generated-router navigation abort superseded work and verify it is still current at every async boundary.

- Adds an incrementing ID and `AbortController` to each navigation; starting a newer navigation aborts the previous request.
- Guards every asynchronous commit point (fetch, plugin, route loading, view transition, rendering) with a check that the navigation is still current.
- Covers both HTML-swap and hydratable production routers, and same-document and document-owned navigation.
- Adds no public API and no cost outside the generated client router.

<sup>Written for commit 005b6d7f4fea055dc3789cb5e19754781b775e1e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/701?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/client-component.test.ts
+++ b/packages/farm/src/__tests__/client-component.test.ts
@@ -736,7 +736,7 @@ export function Chart() {}
     expect(source).toContain("delete window.__farmDocsRuntime;");
     expect(source).toContain("script.replaceWith(freshScript);");
     expect(source).toContain('document.documentElement.dataset.farmDocsRuntime === "true"');
-    expect(source).toContain("this.swapContent(html, url.pathname + url.search)");
+    expect(source).toContain("const swapped = await this.swapContent(");
     expect(source).toContain(
       "const targetUrl = new URL(targetPath || window.location.href, window.location.origin);",
     );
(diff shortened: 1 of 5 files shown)


Kinfe123:
Addressed all three findings. Document-owned and same-document exits now cancel the matching client-plugin session; rendered-path tracking advances at the DOM/history commit instead of after async resolved hooks; and the async swap path carries navigation ownership through lazy hydration work, including isolated boundaries, before later React commits. Added explicit client-plugin cancellation coverage and verified a hydratable production build.

PR #702, opened 2026-09-03
fix(navigation): refresh hydratable production routes

Kinfe123:
## Problem

`router.refresh()` works in the shared router and the production runtime without client components, but the hydratable production router has no refresh method. Calling it through the public navigation API is therefore a no-op, and navigating to the current URL reuses client or prefetched state instead of asking the server for fresh output.

## Root cause

The hydratable runtime was missing the refresh facade and treated every same-path navigation as skippable. Its client-render and intercepted-route fast paths also prevented an explicit refresh from reaching server-rendered HTML, while `fetchPage` had no cache-bypass mode.

## Change

Add the same refresh contract used by the other runtimes. An explicit refresh preserves the document and scroll by default, replaces the current history entry, skips local rendering/interception shortcuts, evicts the matching prefetch entry, and swaps fresh server HTML.

## Safety and compatibility

- Changes behavior only for explicit `refresh: true` / `router.refresh()` calls.
- Normal client-render navigation, route interception, prefetching, hydration, and history behavior are unchanged.
- No public type or API addition is required; this implements the API already exposed by Farm's router facade.

## Verification

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/client-component.test.ts src/__tests__/navigation-router-methods.test.tsx src/__tests__/navigation-state.test.tsx` (47 tests passed)
- `pnpm exec oxlint packages/farm/src/nitro/universal-build.ts packages/farm/src/__tests__/client-component.test.ts` (0 errors; one pre-existing unused-variable warning in `universal-build.ts`)
- `git diff --check`

The generated-runtime regression assertion fails without the hydratable refresh implementation and fresh-fetch path.

## Documentation and release

None. The existing router refresh documentation now applies consistently to production hydration builds.

---
## Summary by cubic
Fixes `router.refresh()` in the hydratable production router so it fetches fresh server HTML instead of reusing client or prefetched state.

- Adds a `refresh` option that bypasses client-render and intercepted-route shortcuts, evicts matching prefetch entries (plain and interception-qualified), and preserves document and scroll while replacing the current history entry.
- Only explicit `refresh: true` or `router.refresh()` calls change behavior; normal navigation, interception, prefetching, and history behavior are unchanged.
- No public API changes; this implements the refresh contract already exposed by Farm's router facade.

<sup>Written for commit c4af2476c54ba180a49a080997a1108b4f777558. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/client-component.test.ts
+++ b/packages/farm/src/__tests__/client-component.test.ts
@@ -757,7 +757,23 @@ export function Chart() {}
     );
     expect(source).toContain('if (error?.name === "FarmDeploymentMismatchError") return;');
     expect(source).toContain("this.fetchPage(pathname, false, false)");
-    expect(source).toContain("this.fetchPage(pathname, interceptFrom, false)");
+    expect(source).toContain("this.fetchPage(pathname, interceptFrom, false, false)");
+  });
+
+  it("refreshes hydratable production routes from uncached server HTML", () => {
+    const source = fs.readFileSync(
+      path.join(process.cwd(), "src", "nitro", "universal-build.ts"),
+      "utf-8",
+    );
+
+    expect(source.match(/refresh: function\(options = \{\}\)/g)).toHaveLength(2);
+    expect(source).toContain(
+      'if (!options.refresh && matched?.route.navigation === "client-render")',
+    );
+    expect(source).toContain("options.refresh ? null : matchInterceptedRouteSlot(pathname, from)");
+    expect(source).toContain("fetchPage: async function(url, interceptFrom, fresh = false");
+    expect(source).toContain("if (fresh) this.clearPrefetchedPath(url)");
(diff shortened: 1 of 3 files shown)


Kinfe123:
Addressed the review finding: a fresh navigation now invalidates both the plain page cache entry and every interception-qualified cache entry for that destination. Added an executable generated-runtime regression test, and reran the focused 40 tests plus the core type-check.

PR #703, opened 2026-09-03
fix(navigation): revert blocked production popstate

Kinfe123:
## Problem

A navigation blocker can reject Back or Forward in the generated production router, but the browser has already traversed to the destination entry before `popstate` fires. The router returns without rendering the destination, leaving the address bar and history position out of sync with the visible page.

## Root cause

The shared SPA router tracks Farm history indexes and reverses blocked traversals. The production-generated routers passed no popstate entry state into navigation and had no rendered-entry index, suppression flag, or reversal path.

## Change

Initialize and maintain a Farm history index in both generated routers, pass popstate state through the navigation attempt, and reverse a rejected traversal with `history.go`. The generated runtime suppresses the resulting return event. A guarded `pushState` fallback restores the rendered URL for legacy entries without usable indexes.

## Safety and compatibility

- Existing history state is preserved when Farm adds its internal index.
- Push, replace, shallow page-state writes, normal navigation, and same-path pop entries keep the index synchronized.
- Allowed popstate behavior is unchanged.
- The fallback matches the shared router for entries created before Farm began tracking indexes.
- No public API changes.

## Verification

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/universal-build-router-runtime.test.ts src/__tests__/client-component.test.ts src/__tests__/spa-router-popstate.test.ts` (42 tests passed)
- `pnpm exec oxlint packages/farm/src/nitro/universal-build.ts packages/farm/src/__tests__/client-component.test.ts packages/farm/src/__tests__/universal-build-router-runtime.test.ts` (0 errors; one pre-existing unused-variable warning in `universal-build.ts`)
- `git diff --check`

The generated-runtime regression test demonstrates that a pop from index 1 to index 0 is reversed and that the return popstate is consumed exactly once.

## Documentation and release

None. This brings production-generated blocker behavior in line with the documented/shared router behavior.

---
## Summary by cubic
Fixes generated production routers so a blocked Back/Forward navigation is reversed instead of leaving the address bar and visible page out of sync.

- Tracks a Farm history index in both generated routers and passes popstate state through the navigation attempt.
- Restores the previous entry with `pushState` when a pop is blocked, with a guarded fallback for legacy entries without usable indexes.
- Existing history state, allowed popstate behavior, and the public API are unchanged.

<sup>Written for commit bdd7d4a2c0d15cc47ccfd8d0fefd7545d9438e3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/farm/src/__tests__/client-component.test.ts
+++ b/packages/farm/src/__tests__/client-component.test.ts
@@ -797,6 +797,10 @@ export function Chart() {}
     ).toHaveLength(2);
     expect(source).toContain("this.observers.set(element, observer);");
     expect(source).toContain("createHistoryState(");
+    expect(source).toContain("revertBlockedPopState: function(from)");
+    expect(source).toContain('if (action === "pop") this.revertBlockedPopState(from)');
+    expect(source.match(/spaRouter\.initializeHistory\(\);/g)).toHaveLength(2);
+    expect(source.match(/void spaRouter\.handlePopState\(event\);/g)).toHaveLength(2);
     expect(source).toContain("currentPath: window.location.pathname + window.location.search");
     expect(source).toContain(
       'if (!options.refresh && action !== "pop" && to === this.currentPath)',
--- a/packages/farm/src/__tests__/universal-build-router-runtime.test.ts
+++ b/packages/farm/src/__tests__/universal-build-router-runtime.test.ts
@@ -99,6 +99,86 @@ describe("generateUniversalRouterStateProperties", () => {
 
     expect([...router.prefetchCache.keys()]).toEqual(["/settings"]);
   });
+
+  it("restores a blocked pop without assuming indexed entries are contiguous", () => {
+    const history = {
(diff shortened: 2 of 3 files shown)


Kinfe123:
Addressed both review findings. History indexes now accept only safe integers, and blocked popstate restoration no longer assumes logical Farm indexes equal physical browser-history distance. It uses a guarded push restoration while preserving the prior Farm state/index. Added executable regressions and reran the focused 40 tests plus core type-check.

PR #708, opened 2026-09-03
fix(navigation): leave blob URLs to the browser

Kinfe123:
## Problem

A `blob:` URL created from the application's origin reports that same origin. Farm's routers therefore classify it as an internal route, attempt page-data or HTML fetches, and intercept raw anchor clicks instead of letting the browser open or download the blob.

## Root cause

The shared and generated routers use origin equality as the complete SPA-eligibility check. Origin is necessary but not sufficient: only same-origin HTTP(S) URLs belong to Farm's route table.

## Change

Centralize URL ownership in a small client helper that requires HTTP(S) and the current origin. Use it for shared navigation/prefetch and both generated production navigation/prefetch paths. Generated click delegates now preserve every explicit URI scheme before calling `preventDefault`, including same-origin `blob:` URLs. The routing guide names this behavior explicitly.

## Safety and compatibility

- Ordinary same-origin HTTP(S) routes remain SPA navigations.
- Cross-origin URLs and non-HTTP schemes remain browser-owned.
- `Link` already preserved explicit URI schemes; raw anchors and direct router calls now agree.
- No renderer-specific code or public API changes.

## Verification

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/spa-router-external-navigation.test.ts src/__tests__/client-component.test.ts src/__tests__/searchparams.test.ts src/__tests__/link.test.tsx` (93 tests passed)
- `pnpm docs:build`
- `pnpm exec oxlint packages/farm/src/client/navigation-url.ts packages/farm/src/client/spa-router.ts packages/farm/src/client/production-runtime.ts packages/farm/src/nitro/universal-build.ts packages/farm/src/__tests__/spa-router-external-navigation.test.ts packages/farm/src/__tests__/client-component.test.ts packages/farm/src/__tests__/searchparams.test.ts` (0 errors; two pre-existing warnings)
- `git diff --check`

The regression tests verify a same-origin blob is external to SPA routing, is never prefetched, and all generated runtime paths use the protocol-aware classifier.

## Documentation and release

Updated the typed-navigation guide to include same-origin `blob:` URLs in browser-owned schemes. No generated artifacts changed.

---
## Summary by cubic
Stops Farm from routing same-origin `blob:` URLs as internal pages, so the browser handles blob navigation and downloads. Navigation and prefetch now accept only same-origin HTTP(S) URLs, while generated click handlers leave all URI schemes and protocol-relative URLs untouched.

**Coverage**
- Adds regression tests for blob URL navigation, prefetching, and digit-bearing custom schemes.
- Documents same-origin `blob:` URLs as browser-owned.

<sup>Written for commit 99b1b75d046ba02835953bba68e117904865c392. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/708?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/src/app/docs/routing/page.md
+++ b/docs/src/app/docs/routing/page.md
@@ -117,7 +117,7 @@ Changing only the fragment preserves SPA state, honors push versus replace histo
 request route data again.
 Native anchor behavior still takes precedence: for example, a `Link` with a `download` attribute is
 handled by the browser instead of Farm's SPA router. Absolute URI schemes such as `mailto:`, `tel:`,
-and `sms:` are also passed through unchanged and are never prefetched as app routes. Literal custom
+`sms:`, and same-origin `blob:` URLs are passed through unchanged and are never prefetched as app routes. Literal custom
 schemes such as `customapp:open` are validated from their URI grammar and work without registration.
 Internal `Link` hrefs stay app-relative: when `basePath: "/console"` is configured, `href="/about"`
 renders and navigates to `/console/about`. Do not add the base path to route hrefs yourself.
--- a/packages/farm/src/__tests__/client-component.test.ts
+++ b/packages/farm/src/__tests__/client-component.test.ts
@@ -797,6 +797,10 @@ export function Chart() {}
       4,
     );
     expect(source).not.toContain("document.querySelector(url.hash)");
+    expect(
+      source.match(/isFarmExternalNavigationURL\(url, window\.location\.origin\)/g),
+    ).toHaveLength(4);
+    expect(source.match(/if \(hasAbsoluteNavigationHref\(href\)\) return;/g)).toHaveLength(2);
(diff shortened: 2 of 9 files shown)


Kinfe123:
Verified the escaping at the emitted-runtime level and removed the ambiguity from the source assertion. The URI grammar now lives in the shared generated runtime helper, and an executable regression proves custom2app:, hyphenated schemes, protocol-relative URLs, and local paths are classified correctly. The four focused files (64 tests) and core type-check pass.

PR #731, opened 2026-09-05
perf(react): compile native reorder pipelines

Kinfe123:
## Problem

Farm can already carry native keyed-array reorder metadata across multiple functional setters queued before one compiler flush. The equivalent expression inside one concise setter still falls back:

```tsx
setRows((current) =>
  current.toSorted((left, right) => left.rank - right.rank).toReversed(),
);
```

That forces complete keyed reconciliation even though every intermediate operation is a native identity-preserving permutation.

## Root cause

The compiler recognizes only a direct `toSorted()` or `toReversed()` call whose receiver is the updater parameter. It does not model a linear native reorder call chain, so the outer call prevents every step from receiving the existing metadata helper.

## Change

- Parse concise functional-updater chains containing two or more `toSorted()` and `toReversed()` calls.
- Lower each method lookup and call in source order through the existing sort/reorder metadata helpers.
- Reuse the existing committed-token propagation and one final LIS reconciliation; no runtime abstraction or public option is added.
- Count each lowered sort/reverse step in the existing compiler report fields.
- Add a 10,000-row production benchmark and compiled fallback control for two cancelling reversals inside one setter.

## Safety and compatibility

The compiler requires compiler-owned, index-independent keyed host rows, a concise updater, noncomputed native method names, and inline synchronous comparators from the existing safe expression subset. A chain containing another method, a referenced comparator, an argument to `toReversed()`, or unsupported updater syntax stays on normal React.

Generated code preserves property lookup, comparator evaluation, native calls, returned values, and thrown errors. The runtime records metadata only for ordinary dense arrays and exact native methods. Custom methods, sparse or subclassed arrays, changed identities, duplicate identities, ambiguous ownership, or failed validation use complete keyed reconciliation before any DOM mutation.

This is React-renderer-only experimental compiler behavior. React 18.3.1 and 19.2.8 compatibility pass. No runtime implementation or feature selection changed; every persisted runtime-size fixture is byte-for-byte unchanged and disabled builds retain no new browser runtime cost.

## Verification

- `pnpm --filter @farm.js/react type-check`
- `pnpm --filter @farm.js/react test` — 645 tests passed; 3 isolated stress tests passed
- `pnpm --filter @farm.js/react test:compat` — React 18.3.1 and 19.2.8 passed
- `pnpm --filter @farm.js/react test:runtime-size` — all persisted budgets passed; 82.6% core-runtime reduction preserved
- `pnpm --filter farm-react-compiler-dashboard-example type-check`
- `FARM_REACT_COMPILER=true FARM_REACTIVITY=hybrid pnpm --filter farm-react-compiler-dashboard-example build`
- `FARM_EXPERIMENT_BROWSER_PATH='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' pnpm --filter farm-react-compiler-dashboard-example benchmark`
  - correctness, general performance, scalability, and every existing optimization gate: PASS
  - pipeline static: 5.60 ms vs React 50.85 ms and compiled control 12.80 ms — 9.08x / 2.29x
  - pipeline hybrid: 5.60 ms vs React 50.85 ms and compiled control 13.00 ms — 9.08x / 2.32x
  - compiler report: 5 reverse steps, including both pipeline calls
- `pnpm --dir docs exec farm generate --check`
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm docs:build`
- focused `oxfmt`, `oxlint`, `node --check`, and `git diff --check`

The runtime differential suite includes 2,000 deterministic updates, each with two to four randomized native sort/reverse steps, compared with normal React. It also covers DOM identity, zero key/descriptor/binding rereads, custom-method fallback, focused input selection, Strict Mode hydration, and unmount-before-flush cleanup.

Environment: Chrome 152.0.7977.82, Node.js 23.11.0, Apple M1 macOS arm64.

## Documentation and release

Updated the React renderer docs, package README, dashboard benchmark guide, executable example, benchmark gate, and persisted benchmark results. No release, template, or generated-route changes.

---
## Summary by cubic
Compiles chained `toSorted()` and `toReversed()` calls inside one concise functional setter, so native reorder pipelines no longer fall back to complete keyed reconciliation.

**Safety and compatibility**
- Unsupported chains, custom methods, sparse/subclassed arrays, and mismatched identities stay on normal React.
- No runtime abstraction or option added; runtime-size fixtures unchanged.
- Each compiled step is reported as `keyedArrayReorderHints` or `keyedArraySortHints`.

**Benchmark and tests**
- New 10,000-row benchmark: 5.60 ms vs 50.85 ms React and 12.80 ms compiled control.
- 645 package tests, stress suite, and React 18.3.1/19.2.8 compatibility pass.
- Docs, README, and benchmark gate updated.

<sup>Written for commit 727550f2eb112b1058c204f81b8d84fe1cb35c29. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/731?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/src/app/docs/renderers/react/page.md
+++ b/docs/src/app/docs/renderers/react/page.md
@@ -1182,13 +1182,14 @@ then applies one LIS-based reconciliation. A cancelling
 `current.toReversed().toReversed()` pipeline therefore performs no DOM moves.
 
 The first proof requires compiler-owned host rows whose render and key do not observe the index.
-Arguments to `toReversed()`, referenced comparators, computed methods, chains containing another
-method, block-bodied updaters, subclassed or sparse behavior, collection-reading bindings, custom
-methods, an unhinted or structural update between reorder setters, React-owned rows, nested host
-blocks, row conditionals, unrelated dirty dependencies, and any identity mismatch use complete
-keyed reconciliation. No option or component is added. Reports count every compiled reverse step
-as a `keyedArrayReorderHints` entry; modules without one do not retain the optional reorder runtime.
-The application runtime must provide `Array.prototype.toReversed`; Farm does not polyfill it.
+Arguments to `toReversed()`, referenced comparators, computed methods, chains containing a method
+other than `toSorted()` or `toReversed()`, block-bodied updaters, subclassed or sparse
+behavior, collection-reading bindings, custom methods, an unhinted or structural update between
+reorder setters, React-owned rows, nested host blocks, row conditionals, unrelated dirty
+dependencies, and any identity mismatch use complete keyed reconciliation. No option or component
+is added. Reports count every compiled reverse step as a `keyedArrayReorderHints` entry; modules
+without one do not retain the optional reorder runtime. The application runtime must provide
+`Array.prototype.toReversed`; Farm does not polyfill it.
 
 #### Keyed array sort hints
 
--- a/docs/src/app/docs/renderers/react/page.md
+++ b/docs/src/app/docs/renderers/react/page.md
@@ -375,10 +375,10 @@ exact retained interval after runtime validation.
 `keyedArrayPositionHints` counts compiler-proven native keyed-array insertions, single or
 contiguous-range removals, single-row replacements, and exact-window replacements with a guarded
 position.
-`keyedArrayReorderHints` counts direct native keyed-array reversals whose complete permutation is
-known at build time.
-`keyedArraySortHints` counts direct native keyed-array sorts whose resulting permutation can be
-validated without rebuilding keyed rows.
+`keyedArrayReorderHints` counts native keyed-array reverse steps whose complete permutation can be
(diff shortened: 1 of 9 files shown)


Kinfe123:
Addressed both review findings in 727550f2: the fallback boundary now explicitly means methods other than native `toSorted()`/`toReversed()`, and the earlier 637-test result is labeled as its historical measurement point rather than the current total.

PR #761, opened 2026-09-07
fix(ci): avoid bundled Corepack in E2E scripts

Kinfe123:
## Problem

The framework E2E job installs pnpm successfully, then its package script launches `corepack pnpm` again. Node 22.12's bundled Corepack cannot verify the current pnpm signing key, so the job exits before any E2E work runs.

## Root cause

Nested package scripts bypass the pnpm binary already installed by `pnpm/action-setup` and re-enter the older Corepack bundled with Node.

## Change

Delegate directly to `pnpm` from every root E2E orchestration script and add a regression check that rejects nested `corepack pnpm` invocations.

## Safety and compatibility

The commands, filters, environment variables, and Playwright suites are unchanged. Only package-manager resolution now follows the caller's installed pnpm binary.

## Verification

- `node --test --test-name-pattern='workspace scripts use' scripts/pnpm-script-shell.test.js`
- `git diff --check`

The complete local script-shell test also exposes an existing empty-variable expectation that differs under the checkout's pnpm 8 shell emulator; that pre-existing assertion is outside this PR.

## Documentation and release

None.

---
## Summary by cubic
Fixes the framework E2E CI job failing on Node 22.12 by replacing `corepack pnpm` with `pnpm` in all root E2E scripts, so the job uses the pnpm binary already installed by `pnpm/action-setup` instead of Corepack's bundled version. Also adds a regression test that rejects `corepack pnpm` in package scripts and Playwright server commands.

<sup>Written for commit 583bb5be93f73b6bb43a829cbe68b2afb527fdc0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/761?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Fixed the review gap in 583bb5be. The regression now scans every root playwright*.config.ts file in addition to package.json scripts, and the three remaining Playwright webServer commands now call the pnpm binary installed by the workflow. The focused regression, lint, and format checks pass.

PR #768, opened 2026-09-07
fix(telemetry): exclude Vercel preview deployments

Kinfe123:
## Problem

Farm's automatic production-site telemetry records public Vercel preview URLs as production
websites. Branch aliases such as
`docs-git-fix-query-array-round-trip-kinfe123s-projects.vercel.app` then appear alongside real
production sites.

## Root cause

The runtime validates and reduces each incoming URL to a public HTTPS origin, but it does not
distinguish Vercel deployment environments. The ingestion service and dashboard consequently accept
every normalized Vercel origin.

## Change

- Skip production-site reporting when `VERCEL_ENV` or `VERCEL_TARGET_ENV` identifies preview,
  development, or a custom non-production environment.
- Keep explicit Vercel production deployments eligible, including production domains ending in
  `.vercel.app`.
- Keep the six Vercel-confirmed legacy preview URLs in one server-owned cleanup definition.
- Hide those exact legacy URLs from dashboard counts immediately, refuse their re-ingestion, and
  remove their stored rows during the ingestion maintenance pass.
- Document the automatic preview behavior and explicit cleanup.

## Safety and compatibility

The general preview decision uses deployment metadata and does not infer preview status from a
hostname. Deployments without Vercel metadata keep the existing behavior, ordinary production
`.vercel.app` domains remain tracked, and a production project slug containing `-git-` remains
eligible. Reporting stays best-effort background work and never blocks application responses.

## Verification

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/product-telemetry.test.ts`
- `./packages/farm/node_modules/.bin/vitest --root docs run src/app/api/telemetry/v1/sites/route.test.ts`
- `pnpm --filter @farm.js/core type-check`
- `pnpm exec oxlint packages/farm/src/product-telemetry.ts packages/farm/src/__tests__/product-telemetry.test.ts docs/src/lib/telemetry-sites.ts docs/src/app/api/telemetry/v1/sites/route.ts docs/src/app/api/telemetry/v1/sites/route.test.ts docs/src/app/telemetry/page.tsx`
- `pnpm exec oxfmt --check packages/farm/src/product-telemetry.ts packages/farm/src/__tests__/product-telemetry.test.ts docs/src/lib/telemetry-sites.ts docs/src/app/api/telemetry/v1/sites/route.ts docs/src/app/api/telemetry/v1/sites/route.test.ts docs/src/app/telemetry/page.tsx docs/src/app/docs/telemetry/page.md`
[...]

PR #771, opened 2026-09-07
fix(ci): avoid rebuilding packages during type checks

Kinfe123:
## Problem

The PR #768 CI run failed in all three full test matrices during the final workspace type-check. The first run reported `src/client.ts(109,10): error TS2451: Cannot redeclare block-scoped variable Link` even though the checked merge ref contains one Link export and focused core/workspace type-checks pass outside the CI graph.

## Root cause

CI already runs `pnpm build` before tests and type-checking. The later `pnpm type-check` command lets Turbo execute parent build tasks again. That causes `@farm.js/core#build` to clean and regenerate declarations while type-check tasks resolve the same package output. Serializing the core tasks only moved the failure into the redundant second build (`TS5055: Cannot write file ... because it would overwrite input file`).

## Change

Add a CI-specific `type-check:ci` script using Turbo `--only`, and use it after the existing successful build step. Normal developer `pnpm type-check` behavior is unchanged. Add a regression assertion that keeps the workflow and script contract together.

## Safety and compatibility

This changes CI orchestration only. CI still builds all packages before running tests and type-checks. Runtime behavior, public APIs, renderers, deployment targets, and package output are unchanged.

## Verification

- `node --test --test-name-pattern="CI type-check does not rerun" scripts/pnpm-script-shell.test.js`
- Turbo dry-run for core and React contains only their type-check tasks and no build tasks
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec turbo type-check --only --force --filter=@farm.js/core --filter=@farm.js/react` (2/2 tasks passed in a clean worktree)
- `pnpm exec oxlint scripts/pnpm-script-shell.test.js`
- `pnpm exec oxfmt --check package.json .github/workflows/ci.yml scripts/pnpm-script-shell.test.js`

## Documentation and release

None.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
           TMP: ${{ runner.temp }}
 
       - name: Type check
-        run: pnpm type-check
+        run: pnpm type-check:ci
         env:
           NODE_OPTIONS: --max-old-space-size=8192
 
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "lint:fix": "oxlint --fix . && oxfmt .",
     "format:check": "oxfmt --check .",
     "type-check": "turbo type-check",
+    "type-check:ci": "turbo type-check --only",
     "clean": "turbo clean",
     "bump": "bumpp",
     "version-packages": "pnpm run bump",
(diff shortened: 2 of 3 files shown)


Kinfe123:
Closing this workaround because the new CI run showed that avoiding Turbo dependency builds did not resolve the Link redeclaration. The reproduced root cause is a production fixture that Nitro evaluates as a root index.mjs, which starts a dev server in the core package and leaves generated source artifacts. The focused fix is #772.

PR #842, opened 2026-09-09
perf(react): preserve row lineage through interleaved maps

Kinfe123:
## Problem

The experimental React compiler can retain same-key replacement lineage when safe maps run before structural keyed-row steps, but a concise pipeline such as `filter().map().slice().toSorted()` drops to complete keyed reconciliation. That fallback re-reads every surviving binding even though the compiler already proved the map, structural membership, and final reorder.

## Root cause

The compiler rejected every map after the first filter or slice. The runtime map recorder also accepted only committed or non-structural reorder sources, while later filter/slice and reorder recorders preferred the older structural mapping instead of the newest flattened replacement lineage.

## Change

- Accept safe maps before or between index-independent filter/slice steps, up to the final native reorder.
- Carry the structural survivor chain and latest mapped-item-to-committed-item mapping through every accepted step.
- Preserve the existing non-structural reorder/map paths and keep maps after a structural reorder on the complete React fallback.
- Update the compiler dashboard workload to benchmark `filter -> map -> reorder`, and document the public boundary and results.
- Add the new 2,000-update differential case to the dedicated stress configuration.

## Safety and compatibility

Every native method and callback still executes in JavaScript order. The runtime validates dense native arrays, the committed collection token, source/result lengths, survivor indices, replacement identity, and every row key before the first DOM write. Changed keys, custom methods, unsupported syntax, no-op runtime structural steps that cannot retain proof, and failed validation use complete keyed reconciliation.

The focused tests preserve surviving DOM nodes, remove only rejected rows, patch only the changed row, keep controlled-input focus and selection, hydrate in Strict Mode, and discard queued work after unmount. React 18.3.1 and 19.2.8 compatibility passed. No public configuration or renderer-neutral API changes.

## Verification

- `pnpm --filter @farm.js/react test` — 734 passed, 10 skipped; stress phase passed.
- `pnpm --filter @farm.js/react test:stress` — 12 heavy tests passed, including both 2,000 mapped-structural differential cases.
- `pnpm --filter @farm.js/react type-check`
- `pnpm --filter @farm.js/react test:compat`
- `pnpm --filter @farm.js/react test:runtime-size`
- `pnpm --filter farm-react-compiler-dashboard-example type-check`
- `pnpm --filter farm-react-compiler-dashboard-example build`
- `pnpm docs:build`
- focused formatting and lint checks for every changed source file

The full Chrome 153 production benchmark passed correctness, the broad 10% regression gate, and the target mapped-structural gate without changing thresholds. At 10,000 rows:

| Mode | Interleaved pipeline | Compiled control | vs React | vs control |
| --- | ---: | ---: | ---: | ---: |
| Static | 5.80 ms | 12.80 ms | 11.28x | 2.21x |
| Hybrid | 5.80 ms | 12.90 ms | 11.28x | 2.22x |

The overall benchmark command reported three unrelated timing-floor misses: append scaling at 3.996x versus 4x, static structural control at 1.239x versus 1.25x, and hybrid multi-map reorder at 3.19x versus 4x. Their fixtures and thresholds are unchanged; the machine-readable report is `/tmp/farm-react-interleaved-map-benchmark.json`.

## Documentation and release

Updated the React package README, React renderer documentation, compiler dashboard README, executable benchmark fixture, and persisted benchmark result. No version or release-flow changes.

---
## Summary by cubic
Preserves keyed row lineage when safe maps run between index-independent filter/slice steps before a final native reorder, so pipelines like `filter().map().slice().toSorted()` keep partial reconciliation instead of dropping to complete keyed re-reads. Maps after the structural pipeline has already reordered, structural work split across setter calls, unsafe callbacks, and failed runtime validation still use the full fallback.

**Details**
- The runtime now carries survivor indices and mapped-item sources through each accepted step, and the compiler accepts maps before or between filter/slice steps.
- Every native method and callback still runs in JavaScript order; survivor indices, keys, lengths, and replacement identity are validated before the first DOM write.
- The dashboard benchmark now exercises `filter -> map -> reorder`, and a 2,000-update differential stress case matches normal React.
- The runtime size budget was bumped to keep the interleaved map support within the size gate.

<sup>Written for commit ccb367a5a3c30fddd08d5f1d16e054be758294c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/842?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/src/app/docs/renderers/react/page.md
+++ b/docs/src/app/docs/renderers/react/page.md
@@ -1280,24 +1280,26 @@ controlled inputs, focus, and text selection stay attached to their keys. Multip
 map-and-reorder setters queued before one compiler flush compose against the same committed
 collection and expose only the final state.
 
-Safe maps can also lead into an index-independent filter or slice before the native reorder:
+Safe maps can also appear before or between index-independent filter and slice steps before the
+final native reorder:
 
 ```tsx
 setItems((current) =>
   current
-    .map((item) => (item.id === editedId ? { ...item, label: nextLabel } : item))
     .filter((item) => item.visible)
+    .map((item) => (item.id === editedId ? { ...item, label: nextLabel } : item))
+    .slice(0, limit)
     .toSorted((left, right) => left.rank - right.rank),
 );
 ```
 
-The filter or slice still evaluates the mapped values in normal JavaScript order. Farm carries both
-proofs to the final commit: which committed rows survived and which surviving items replaced their
-source items. It validates every survivor and key before the first DOM write, removes rejected
-rows, performs only the final LIS moves, and patches bindings only for changed survivors. A changed
-item that the structural step removes needs no row patch. This combined path is limited to one
-concise functional setter with maps before the structural steps; maps after a filter/slice and
-structural work split across setter calls keep complete reconciliation.
+Each callback still evaluates in normal JavaScript order. Farm carries both proofs across the
+pipeline: which committed rows survived and which surviving items replaced their source items. It
+validates every survivor and key before the first DOM write, removes rejected rows, performs only
+the final LIS moves, and patches bindings only for changed survivors. A changed item that a later
+structural step removes needs no row patch. This combined path is limited to one concise functional
+setter. A map after the structural pipeline has reordered, structural work split across setter
+calls, or any failed runtime proof keeps complete reconciliation.
 
 A safe map may also be the final pipeline step:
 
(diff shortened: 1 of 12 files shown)


Kinfe123:
Runtime-size follow-up for the Linux React 19 check: the keyed map/reorder fixture measures 13,470 B gzip against the persisted 13,143 B baseline. This PR gives that feature a 328 B allowance (13,471 B maximum), so the gate retains one byte of headroom. The filter-only platform variance is bounded at 12,465 B maximum. The mapped-reorder cleanup also reduced the local fixture by 2 B gzip; correctness and performance thresholds are unchanged.

PR #958, opened 2026-09-12
feat(api): create server and HTTP callers from one setup

Kinfe123:
## Problem

Apps should define their API callers once, in one shared module, instead of maintaining separate server and browser setup. The existing HTTP factory and explicit endpoint-map server factory do not provide that contract. The API-client docs also incorrectly showed `{ request }` as enough for automatic route discovery.

Follow-up to #951.

## Root cause

The HTTP caller already owns typed route proxies, scoped dynamic params, validation-result decoding, and caching. The old server factory simply returns the endpoint functions supplied by the app; it has no connection to the current app's registered route dispatcher.

The new end-to-end test also exposed an existing dev refresh gap: changing a file endpoint regenerated types but did not refresh the API handler registry.

## Change

```ts
// src/lib/api.ts — one shared definition
import { createApiClients } from "@farm.js/core/client";
import { apiRoutes, type APIRouter } from "./api.generated";

export const { api, apiClient } = createApiClients<APIRouter>({ routes: apiRoutes });
```

- `api` resolves the current request and owning app's dispatcher and invokes local endpoints without `fetch`.
- `apiClient` uses the existing HTTP implementation. Both app-route callers return `{ data, error, key }` with the same generated input/output/method inference and `$params()` scopes.
- Reuse the existing proxy, transport encoding/decoding, and cache implementation. Keep server caches and in-flight work isolated by request, app, and credentials, without process-wide invalidation subscriptions retaining request data.
- Bind the dispatcher in core development, standalone API plugins, universal production output, and the RSC request entry. Keep handlers and Node imports behind a browser-safe bridge.
- Refresh file endpoint handlers after Vite invalidation so a shared server caller sees HMR updates.
- Migrate docs, maintained examples, and renderer starter templates to the shared pair, including generated path/method manifests by default.

## Safety and compatibility

- This is direct endpoint dispatch, **not a replay of the full HTTP pipeline**. Route matching, endpoint middleware, input/output validation, body limits, and response decoding still apply. Path-level HTTP middleware, redirects/rewrites, plugin request/response lifecycle hooks, and deployment checks are not rerun. Authorization shared by both transports belongs in endpoint middleware.
- Local calls still serialize request/response data; this removes the network round trip, not all serialization. No benchmark or broader performance claim is made.
- The active request's Cookie, Authorization, and Accept-Language are inherited, with explicit header overrides supported. The local API mount is authoritative; a public external `baseURL` cannot redirect app-route `api` calls or receive their cookies.
- Module-level construction is safe. Calling app-route `api` in a browser or outside a Farm request fails clearly instead of falling back to HTTP. Response cookies/headers do not automatically propagate to the enclosing response; request cancellation and the enclosing after lifecycle are retained.
- Existing `createAPIClient`, `createServerAPIClient`, and integration factories remain supported. Integration aliases retain their existing dispatch semantics and shared options.
- Core behavior is renderer-neutral. The documented production path is the default universal runtime; legacy `universal: false` is not given automatic local dispatch.

## Verification

Local environment: macOS, Node 23.11.0, pnpm 8.12.1. Windows and other Node versions remain for CI.

- `NODE_OPTIONS=--max-old-space-size=6144 pnpm --filter @farm.js/core build` — passed, including declarations; the 6 GB heap matches the workspace release build setting.
- `pnpm --filter @farm.js/core test -- src/__tests__/api-client.test.ts src/__tests__/api-client-scopes.test.ts src/__tests__/api-client-transport.test.ts src/__tests__/api-client.typing.test.ts src/__tests__/api-clients.test.ts src/__tests__/api-route-manager.test.ts src/__tests__/api-vite-plugin-hmr.test.ts src/__tests__/api-vite-plugin-rewrite.test.ts src/__tests__/api-vite-plugin-collision-hmr.test.ts src/__tests__/type-generator.test.ts src/__tests__/endpoint-middleware.test.ts src/__tests__/client-cache.test.ts src/__tests__/request-boundary.test.ts` — 138 tests passed before the additional integration-alias case.
- `pnpm --filter @farm.js/core test -- src/__tests__/api-clients.test.ts src/__tests__/integration-client.test.ts` — 38 tests passed, including the final 11 paired-caller tests.
- `pnpm --filter @farm.js/core test -- src/__tests__/plugin-routes-production.test.ts` — passed against a real dev server and built Vercel output. Checks SSR, nested local calls, dynamic plugin routes, custom API mounts, validation/auth failures, error sanitization, generated type inference, and server-only sentinel exclusion from browser JS. Production local calls and SSR succeed with global `fetch` deliberately disabled. The HMR assertion failed before the handler-refresh fix and passes with it.
- `pnpm --filter @farm.js/plugin build` — passed.
- `pnpm --filter @farm.js/plugin test -- src/rsc/entries/security.test.ts src/rsc/entries/after.test.ts src/rsc/entries/api-route-shapes.test.ts src/rsc/entries/page-routing.test.ts` — 32 tests passed, including executable request-context coverage.
- `pnpm --filter @farm.js/create-app test` — all 16 generator tests passed.
- `pnpm docs:check-navigation`, `pnpm --dir docs exec farm generate --check`, and `pnpm docs:build` — passed. The docs build produced Vercel output locally; nothing was deployed.
- Focused `pnpm exec oxfmt` / `oxlint` and `git diff --check` — passed; oxlint reports only the eight existing unused-variable warnings in touched runtime/plugin files.

## Documentation and release

Updated API-client, API-route, getting-started, configuration, plugin lifecycle, preview, and server-query guidance, plus the framework authoring skill. Migrated the docs app, basic and renderer examples, StackBlitz example, and renderer starter templates. Regenerated their API manifests and added generator assertions. UI styling is unchanged.

No dependency upgrades, package version changes, tags, or release actions.

---
## Summary by cubic
Introduces `createApiClients()`, a single `src/lib/api.ts` setup that returns `{ api, apiClient }` and replaces separate server and browser client factories. `api` dispatches directly to the current request's registered endpoint without a network round trip; `apiClient` keeps the existing HTTP implementation.

- Both callers share typed route inference, validation, caching, and the `{ data, error, key }` result shape.
- Direct `api` calls run endpoint middleware and input/output validation but skip outer HTTP middleware, redirects, and plugin lifecycle hooks.
- The dispatcher is bound in dev, standalone API plugins, universal production output, and the RSC request entry; server caches and in-flight work stay scoped to request, app, and credentials.
- Adds HMR refresh for file endpoint handlers after Vite invalidation.
- Existing `createAPIClient` and `createServerAPIClient` remain supported; `universal: false` gets no local dispatch.
- Updates docs, examples, and starter templates to use the shared caller, and generated manifests are now emitted by default.

<sup>Written for commit e29b3fc48b312692309be78243237660ec733ba1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
CI follow-up: updated the RSC search-params source assertion to include the new API request-context import. The repeated-key checks are unchanged; no tests were skipped or timeouts widened. The full plugin suite now passes locally: `pnpm --filter @farm.js/plugin test` — 22 files, 127 tests, including the production RSC build/boot fixtures. Also refreshed the branch against current main (including the beta.94 release); CI is rerunning on the updated head.

PR #962, opened 2026-09-13
docs(api): clarify shared route and integration callers

Kinfe123:
## Problem

Follow-up to #958. The API-client guide recommends `createApiClients()` but its integration example still calls `apiClient.billing`, which belongs to the integration-only factory. The integrations guide also presents a second caller setup without explaining when it is needed, and the scoped-route section still describes the paired server caller as HTTP-only.

## Root cause

The new factory reuses the existing integration namespace under `.integrations`, while `createIntegrations()` exposes configured names directly. The implementation already supports both routes and integrations in one pair, but the prose and examples mixed the two contracts. Runtime tests do not check the Markdown guidance.

## Change

- Show one `createApiClients<APIRouter, AppIntegrations>({ routes: apiRoutes })` setup for app/plugin routes and integrations.
- Correct integration calls to `api.integrations.billing` and `apiClient.integrations.billing` in the paired-factory guide.
- Show the server registry, its `farm.config.ts` registration, and the shared caller module together. Explain which inputs are TypeScript-only and which contain public runtime metadata.
- Document integration-only caller paths as a supported alternative, not a required second setup.
- Explain plugin versus integration ownership, browser-safe options, server-only operations, and the existing differences in results and HTTP fallback.
- Correct the stale scoped-route server paragraph and update the Farm authoring skill to recommend the same shared setup.

## Safety and compatibility

Documentation only. No runtime, public API, provider registration, renderer, target, or generated declaration changes. `createIntegrations()` remains supported and is not deprecated. The paired factory does not accept its explicit source map or separate server-options argument. Provider clients and credentials stay server-only; the shared module imports only registry types and schema-free route metadata.

## Verification

Local environment: macOS arm64, Node 23.11.0, pnpm 8.12.1.

- `pnpm --filter @farm.js/core exec vitest run src/__tests__/api-clients.test.ts src/__tests__/api-client.typing.test.ts src/__tests__/integration-client.test.ts` — 44 tests passed.
- `pnpm docs:check-navigation` — passed; 83 visible pages and 12 official plugins.
- `pnpm --dir docs exec farm generate --check` — passed.
- `pnpm exec oxfmt --check docs/src/app/docs/api-client/page.md docs/src/app/docs/integrations/page.md docs/src/app/docs/integrations/custom/page.md skills/farmjs/SKILL.md` — passed.
- `git diff --check` — passed.
- `pnpm docs:build` — initial Vercel production build passed. The built API-client page was inspected in a local browser. Final explanatory prose additions were checked with formatting, navigation, and generated-type checks; final-commit production verification is left to PR CI.

No regression test was added because no runtime mechanism changes. Existing paired-caller tests verify the integration aliases and shared options on both callers. This change does not alter layout, styling, or interactions.

## Documentation and release

Updates the API-client guide, integrations overview, custom-integration guide, and `skills/farmjs/SKILL.md`. No release bump, templates, package changes, or generated artifacts.

Suggested separate follow-up: propagate request cancellation through local integration dispatch, with tests for already-aborted requests and in-flight cancellation. Do not change provider ownership or existing HTTP-fallback compatibility in this docs PR.

---
## Summary by cubic
Clarifies the Farm API-client docs so app/route and integration callers share one `createApiClients()` setup instead of the guide mixing in a separate `createIntegrations()` factory. Also covers when the integration-only factory is the better choice. Docs-only; no runtime, package, or generated-declaration changes.

- Integration calls in the shared setup use the reserved `api.integrations.billing` / `apiClient.integrations.billing` namespace; the integration-only factory still exposes `api.billing` directly and is not deprecated.
- Explains which inputs (`APIRouter`, `AppIntegrations`) are TypeScript-only and which (`apiRoutes`, `integrations.data`) are runtime metadata, and keeps provider SDKs server-only.
- Documents that app-route results keep `{ data, error, key }` while integrations keep `{ data, error }` and server-side HTTP fallback; the paired server caller is not HTTP-only.
- Updates the integrations overview, custom-integration guide, and `skills/farmjs/SKILL.md` to recommend the shared setup, with `createIntegrations()` covered as an integration-only alternative.
- Switching factories is not a rename: add `.integrations` and move shared integration defaults into the `integrations` option; `integrations: false` only disables that caller namespace, not registration or routes.
- Adds an options comparison so `createIntegrations()` stays preferred when setup-level `request`, `forwardHeaders`, `serverOptions`, or explicit source maps are needed; the shared factory accepts those defaults under `integrations` and per-call server overrides instead.

<sup>Written for commit 18051da1afb93cbf35869d64ef85d2c7735a7e48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/962?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/src/app/docs/api-client/page.md
+++ b/docs/src/app/docs/api-client/page.md
@@ -581,14 +581,21 @@ never fall back to HTTP. Operations marked `isServer: true` remain available onl
 
 ### Integration-only callers
 
-`createIntegrations<AppIntegrations>()` remains supported when only integration callers are
-needed. It returns integration namespaces directly: `apiClient.billing.checkout.post(...)` and
-`api.billing.checkout.post(...)`. With `createApiClients`, those same calls need the
+`createIntegrations<AppIntegrations>()` remains supported and is not deprecated. Existing apps
+do not need to migrate. Use it when only integration callers are needed, or when you prefer to
+keep them separate from app-route callers. It returns integration namespaces directly:
+`apiClient.billing.checkout.post(...)` and `api.billing.checkout.post(...)`.
+With `createApiClients`, those same calls need the
 `.integrations` segment. Choose one setup for the shared module; do not create both pairs for
 the same integrations. Switching factories requires updating the namespace and moving shared
 integration defaults into the `integrations` option; it is not a drop-in rename.
 The paired factory discovers configured integrations; it does not accept the integration-only
 factory's explicit source map or separate server-options argument.
+
+For deliberately separate modules, use `createApiClients<APIRouter>({ routes: apiRoutes,
+integrations: false })` for app routes and `createIntegrations<AppIntegrations>()` for integration
+callers. Disabling the paired factory's namespace does not unregister integrations or their
+HTTP routes; it only leaves integration access to the separate caller module.
 
 ## Server Function Form Actions
 
--- a/docs/src/app/docs/api-client/page.md
+++ b/docs/src/app/docs/api-client/page.md
@@ -26,6 +26,9 @@ paths/methods and types, never server handlers or credentials. Import `api` in s
 Both app-route callers return `{ data, error, key }` and preserve the same input, output, method,
 and dynamic-parameter inference. Keep secrets out of this shared module, including its options.
 
+If the app also has integrations, add `AppIntegrations` as the second type argument to this same
+factory. You do not need a second caller setup. See [Integration callers](#integration-callers).
+
 The HTTP client uses the current origin and `/api` by default. To point every default client at another
 API, configure it once in `farm.config.ts`:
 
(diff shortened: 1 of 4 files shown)


Kinfe123:
Added an explicit compatibility note: createIntegrations() is still supported, is not deprecated, and does not require migration. Apps may keep route and integration callers in separate modules by disabling the combined factory’s integrations namespace; the configured integrations and their HTTP routes remain registered. Updated both guides and the Farm authoring skill. Verification: 38 focused caller tests, formatting, docs navigation, and generated-type checks passed.

Kinfe123:
Expanded the compatibility callout with an options comparison and a request-scoped server example. Shared baseURL, headers, credentials, and data defaults work in either factory; setup-level request/forwardHeaders, separate server defaults, and explicit source maps are concrete reasons to retain createIntegrations(). The docs also distinguish setup defaults from per-call overrides and warn against exposing server options or caching request-bound callers globally. Updated authoring guidance as well. The 38 focused caller tests and formatting/navigation/generated-type checks passed.

PR #965, opened 2026-09-13
feat(client): resolve instance headers per call

Kinfe123:
## Problem

Caller instances accept static header records, but cannot derive fresh synchronous or asynchronous defaults at call time. Apps otherwise have to thread those values through individual calls.

```ts
export const { api, apiClient } = createApiClients<APIRouter>({
  routes: apiRoutes,
  headers: async () => ({ "Accept-Language": await readPreferredLanguage() }),
});
```

The same `headers` option should work with `createIntegrations`, including separate server defaults and explicit-source callers.

## Root cause

The HTTP callers treat `options.headers` as a record; paired server route callers additionally capture it on their first call within a request. Passing a function is either ignored or fails header construction. Resolving only at fetch time would also leave route cache identity, QUERY keys, and retry behavior inconsistent with the sent headers.

## Change

- Add the shared `ClientHeaders` type: a string-valued record or a sync/async zero-argument resolver returning one.
- Snapshot defaults once per operation call, before route cache lookup or integration dispatch. Synchronous defaults retain a synchronous resolution path.
- Use that snapshot for route cache identity, QUERY keys, optimistic/invalidation target keys, dispatch, and retries. Keep forwarded server request headers separate from caller defaults so paired `api` calls resolve afresh.
- Merge custom integration headers case-insensitively in the browser, matching server precedence. Retain Farm's integration protocol marker on both paths.
- Return resolver errors through the existing result lifecycle without HTTP, direct dispatch, cached data, or optimistic updates.
- Update the published client declaration facade as well as source types. Add runtime and TypeScript regression coverage.

## Safety and compatibility

- Static header objects remain supported. Per-call headers remain objects and override instance defaults; integration operation headers sit between instance defaults and per-call overrides.
- `integrations.headers` replaces shared defaults for integration calls only. A separate integration server `headers` option replaces the client defaults for `api`, preserving existing option precedence.
- Every call resolves headers, including cache hits. Retries reuse the resolved snapshot; changing effective headers does not reuse another header context's cache/in-flight request.
- No new request forwarding, timeouts, retries, credentials, or provider imports. Resolvers are not a server-only/bundler boundary, and secrets must stay out of shared caller modules.
- Renderer-neutral implementation. Real browser checks use the React basic example, in Vite development and built Nitro node-server production. Other renderers/targets are not claimed as separately exercised.

## Verification

macOS arm64, Node 23.11.0, pnpm 8.12.1:

- Against unchanged `main` (`b8591d6e`) in an isolated worktree, all 15 tests in `client-header-resolvers.test.ts` fail; they pass with this change.
- `pnpm --filter @farm.js/core test src/__tests__/client-header-resolvers.test.ts src/__tests__/client-headers.typing.test.ts src/__tests__/api-client.test.ts src/__tests__/api-clients.test.ts src/__tests__/integration-client.test.ts src/__tests__/api-client.typing.test.ts` — 105 passed.
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm --filter @farm.js/core build` — ESM, CJS, and declarations. The first attempt with the default Node heap exhausted the declaration worker's memory; the explicit heap build passes.
- `pnpm --filter @farm.js/cli build`
- `pnpm exec playwright test tests/e2e/framework-features.spec.ts --grep 'runs integration handlers'` — desktop Chrome, 1 passed.
- `pnpm exec playwright test tests/e2e/framework-features.spec.ts --grep 'runs integration handlers' --config playwright.production.config.ts` — Chromium, 1 passed. The test checks server defaults plus French → Spanish browser defaults without recreating the instance. Installed the missing Playwright headless browser locally; no test/config weakening.
- `pnpm --dir docs exec farm generate --check`
- `pnpm docs:build` — Vercel production output passes.
- Focused `pnpm exec oxfmt --check` and `pnpm exec oxlint` on changed supported files; lint has no errors and one existing unused `resolveSourceAPI` warning. Example formatting follows the existing style (examples are excluded by repository formatter config).
- `git diff --check`

## Documentation and release

Document resolver timing, override precedence, error handling, cache/retry behavior, and browser/secret boundaries in the API client and integrations guides. Update the Farm.js skill and existing integration-lab example/browser test. No new page, generated route contract, dependency, or release-version changes.

This is a separate feature PR from #962; it does not include that PR's unified-caller documentation changes.

---
## Summary by cubic
Callers' instance `headers` previously had to be a static record; they now also accept a sync or async resolver and are re-read once per operation call instead of being captured up front. The resolved header snapshot feeds cache identity, QUERY keys, optimistic/invalidation targets, dispatch, and retries, so execution state stays aligned with the sent headers.

- Supported in `createApiClients`, `createAPIClient`, `createIntegrations`, and their explicit/server overload forms; `@farm.js/core/client` now exports the `ClientHeaders` type.
- Header precedence is forwarded request headers, then instance defaults, then integration operation headers, then per-call `headers`, with case-insensitive merging on both browser and server.
- `integrations.headers` replaces shared defaults for integration calls only; the second `createIntegrations` server argument replaces the defaults for `api` instead of merging.
- Resolver errors return the normal error result without fetching, dispatching, reading cached data, or applying optimistic updates.
- Every operation resolves headers, including cache hits; changed effective headers no longer reuse the previous cache entry or in-flight request, and retries reuse only the first snapshot.
- Static header objects and per-call header objects keep working, so no code changes are needed for existing callers; keep secrets out of shared modules because resolvers are not a server-only boundary.
- The published-declaration typing test also passes on Windows via normalized virtual compiler paths.

<sup>Written for commit d903663e17f1ad0c48bd5f46dfc15dae21e8f590. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/965?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/src/app/docs/api-client/page.md
+++ b/docs/src/app/docs/api-client/page.md
@@ -58,6 +58,52 @@ export const { api, apiClient } = createApiClients<APIRouter>({
 Farm forwards `credentials` to every route request from that client. The API must also allow the
 calling origin and credentialed requests through its CORS policy.
 
+## Header defaults
+
+Instance `headers` accepts a string-valued object or a function returning one. The function can
+be synchronous or asynchronous; Farm calls it when an operation is called, not when the client
+is created.
+
+```ts
+import { createApiClients } from "@farm.js/core/client";
+import { apiRoutes, type APIRouter } from "./api.generated";
+
+export const { api, apiClient } = createApiClients<APIRouter>({
+  routes: apiRoutes,
+  headers: () => ({
+    "Accept-Language":
+      typeof document === "undefined" ? "en" : document.documentElement.lang || "en",
+  }),
+});
+
+// This call overrides the instance language without changing later calls.
+await apiClient.hello.get({ headers: { "accept-language": "fr" } });
+```
+
+Use `headers: async () => ({ ... })` when obtaining defaults requires asynchronous work.
+The exported `ClientHeaders` type describes all three forms. Static objects remain supported;
+per-call `headers` remain objects, not resolver functions.
+
+Header names are case-insensitive. App routes merge forwarded server request headers first,
+then instance defaults, then per-call headers. Thus the example explicitly chooses English on
+the server; omit its `Accept-Language` default when you want `api` to preserve the incoming
+request's language. The existing server forwarding allowlist is unchanged.
+
+Farm snapshots the resolved defaults once per operation call, before cache lookup or dispatch.
+Cache hits still resolve headers; changed effective headers isolate the route client's cache
+and in-flight deduplication. Retries reuse the same snapshot. A thrown error, rejected promise,
+or invalid header produces the normal error result without fetching or returning cached data.
+
+These defaults also apply to `api.integrations` and `apiClient.integrations` when enabled.
+Providing `integrations.headers` replaces the shared header defaults for integration calls only;
+it does not merge the two resolvers. See [integration header defaults](/docs/integrations#header-defaults)
+for integration-specific precedence and separate server defaults.
+
+A resolver is not a server-only boundary. Anything imported into this shared module may reach
+the browser, and the resolver runs wherever its caller runs. Do not put provider credentials or
+private environment values here. Read request-specific values inside the resolver or a
+request-scoped server module, never by mutating a shared singleton with one user's credentials.
+
 ## Call a route
 
 **Browser usage**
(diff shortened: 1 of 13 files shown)


Kinfe123:
Fixed the Windows CI failure in d903663e. TypeScript normalized the virtual root filename to forward slashes, while the test host compared it with the native backslash path, so TS6053 reported the fixture missing. The fixture path is now normalized before compiler registration/comparison. This is test-only; no runtime change, platform skip, or timeout increase. Both focused suites pass locally (16 tests), and the full CI matrix is rerunning before merge.

PR #977, opened 2026-09-13
fix(api): preserve invalidations during in-flight reads

Kinfe123:
## Problem

A cached GET that started before a mutation can finish after invalidation and mark its old response fresh again. A subsequent read then reuses stale data.

## Root cause

Cache writes compare invalidation timestamps against response completion time. This loses request ordering, including invalidations in the same millisecond and invalidations cleared by a newer shared-cache writer.

## Change

Observe key-specific invalidation events while an API transport is in flight. Its result still returns to the original caller, but cannot populate the cache after invalidation. Remove the observer in `finally`. Keep the existing newer-request ownership check.

## Safety and compatibility

Renderer-neutral shared API runtime; no new public option, retry, forced request, or periodic timer. Public/private caches remain isolated as before. Requests started after invalidation can cache normally; unrelated keys do not suppress writes. Existing cache subscribers may ignore the internal event argument. Alias invalidation propagates to canonical-key observers.

## Verification

macOS arm64, Node 23.11.0:

- Before the fix: `pnpm exec vitest run src/__tests__/api-inflight-invalidation.test.ts` fails five regressions; the unaffected-key control passes.
- After the fix, from `packages/farm`: `pnpm exec vitest run src/__tests__/api-inflight-invalidation.test.ts src/__tests__/api-client.test.ts src/__tests__/api-clients.test.ts src/__tests__/client-cache.test.ts src/__tests__/server-query-client.test.tsx` — 74 passed.
- `pnpm --filter @farm.js/core type-check` — passed.
- Focused `pnpm exec oxfmt` and `pnpm exec oxlint` — passed.

The tests use deferred responses and controlled clocks, covering private/shared caches, same-tick invalidation, and another client writing the fresh response. No timing sleeps.

## Documentation and release

Documented in the API-client caching guide. No release/version changes.

---
## Summary by cubic
Fixes a race condition where a cached GET started before invalidation could finish after it and re-mark its stale response as fresh. In-flight requests now observe key-specific invalidation events and skip cache writes when invalidated; the result still returns to the original caller.

- Replaces timestamp comparison, which missed same-millisecond invalidations and those cleared by a newer shared-cache writer.
- Preserves the existing newer-request ownership check; requests started after invalidation cache normally, and unrelated keys are unaffected.
- Cache listeners now receive an optional `invalidate` event argument that existing subscribers can ignore; alias invalidation propagates to canonical-key observers.
[...]

PR #978, opened 2026-09-13
fix(cache): release abandoned private client subscriptions

Kinfe123:
## Problem

Creating and abandoning private API callers leaves their caches strongly reachable through the global invalidation listener Set. Even expired data remains reachable until another read performs lazy expiration.

## Root cause

The cache constructor subscribes with a closure capturing `this`. Context-change disposal works for a retained caller, but there is no lifecycle event when the caller itself becomes unreachable.

## Change

Use weak cache subscriptions when the runtime supports `WeakRef`. A finalizer removes abandoned subscriptions; the next invalidation also removes a dead reference when finalization has not run. Explicit disposal unregisters the finalizer. The subscription closure only captures the weak reference, not the cache.

## Safety and compatibility

No public disposal API or reserved caller property. Live caches still receive invalidation, identity isolation is unchanged, and request-local caches remain unsubscribed. Runtimes without `WeakRef` retain the existing strong subscription rather than losing invalidation or failing startup. Garbage collection/finalization timing is not a correctness dependency or promise.

## Verification

macOS arm64:

- Before fix: `pnpm exec vitest run src/__tests__/client-cache-lifetime.test.ts` — three new lifecycle assertions fail, two controls pass.
- After fix, from `packages/farm`: `pnpm exec vitest run src/__tests__/client-cache-lifetime.test.ts src/__tests__/client-cache.test.ts src/__tests__/api-client.test.ts src/__tests__/api-clients.test.ts src/__tests__/client-fetch.test.ts` — 74 passed on Node 23.11.0.
- Additional real-GC smoke: bundled the source API client with esbuild, created ten authenticated cached callers, dropped callers, ran Node 24.21.0 with `--expose-gc` across event-loop turns, then invalidated an unrelated key. Added listener count changed from 10 to 0; all ten WeakRefs were cleared. This is a local smoke check, not a timing-dependent CI test or heap-size benchmark.

Committed tests simulate weak-reference/finalizer events deterministically and cover live invalidation, cleanup without finalizers, explicit/idempotent disposal, request-local opt-out, and the no-WeakRef fallback.

## Documentation and release

API-client guide explains cache ownership, lazy entry expiration, cleanup timing, and the older-runtime fallback. No version changes.

---
## Summary by cubic
Fixes abandoned private client caches staying strongly reachable through the global invalidation listener. On runtimes that support `WeakRef`, cache subscriptions now hold a weak reference instead of the cache, so unreachable callers can be collected instead of lingering until lazy expiration.

**Details**
- Live caches still receive invalidations, and request-local caches still skip global subscriptions.
- Explicit disposal is idempotent and unregisters the finalizer.
- If finalization hasn't run, the next invalidation removes the dead subscription.
- Runtimes without `WeakRef` keep the existing strong subscription, so invalidation still works.
- The API client docs now describe cache lifetime and the older-runtime fallback.
- Adds deterministic tests for finalizer cleanup, no-finalizer cleanup, disposal, request-local opt-out, and the no-`WeakRef` fallback.

<sup>Written for commit 46b20c3859fb88f864c14807ac64c1a62e5c3bb8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:

PR #979, opened 2026-09-13
fix(query): refresh stale data when queries are re-enabled

Kinfe123:
## Problem

After `useServerQuery` loads once, toggling `enabled` off and back on with the same input leaves expired or missing data unrefreshed.

## Root cause

The hook remembers that the current key was already read. Re-enabling runs the effect, but neither its first-read condition nor the explicit-invalidation condition matches ordinary TTL expiration or a cleared cache.

## Change

Reset the remembered read key while automatic reads are disabled. Re-enabling uses the existing first-read path and shared fetch/deduplication logic.

## Safety and compatibility

React `useServerQuery` only; no renderer-neutral routing or server protocol changes. Fresh data is reused, pending work is joined, and disabling does not cancel already-running work. Zero `staleTime` does not create a refetch loop. Manual refetch remains available while disabled. No polling, new public options, or changes to server boundaries.

## Verification

macOS arm64, Node 23.11.0, jsdom/React:

- Before fix: `pnpm exec vitest run src/__tests__/server-query-client.test.tsx` — three new regressions fail, eight tests pass.
- After fix, from `packages/farm`: `pnpm exec vitest run src/__tests__/server-query-client.test.tsx src/__tests__/server-query.test.ts src/__tests__/server-query-boundary.test.ts src/__tests__/client-cache.test.ts` — 31 passed.

Reproducible React render tests cover staleTime zero/positive, initially disabled reads, fresh-cache reuse, cache removal while disabled, and an in-flight request across toggles. No arbitrary sleeps.

## Documentation and release

Documented enabled/disabled/re-enabled behavior in the server-query guide. No version changes.

---
## Summary by cubic
Fixes `useServerQuery` so re-enabling a disabled query refreshes stale or missing data instead of leaving it unread. The hook now forgets the last read key while `enabled` is false, so re-enabling uses the existing first-read path and shared fetch/deduplication.

**Behavior**
- Disabling does not cancel an already-running read; re-enabling joins that pending request.
- Fresh cache is reused, and `staleTime: 0` does not cause a refetch loop.
- Manual `refetch()` remains available while automatic reads are disabled.
- This is limited to the React `useServerQuery` hook; no server protocol or public API changes.
- Adds regression tests and updates the server-queries docs.

<sup>Written for commit a09fa135eb35bf97e7e7c0231676ccb1c3ea2cae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/979?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:

PR #980, opened 2026-09-13
fix(query): preserve prototype-named search parameters

Kinfe123:
## Problem

Query parameters such as `?toString=one` or `?hasOwnProperty=two` become arrays containing inherited JavaScript functions instead of strings. JSON serialization then turns the function entries into null.

## Root cause

The shared search-parameter helper reads a plain object's inherited properties while deciding whether a query key has already appeared.

## Change

Only inspect an existing value when the result owns the property. Keep the ordinary object prototype and the existing blocked-key policy.

## Safety and compatibility

The same helper serves development rendering, production SSR, page-data navigation, hydration, and workflow request queries. No new runtime branch per renderer or public API. Repeated values, empty strings, insertion order, URL decoding, and filtering of `__proto__`, `constructor`, and `prototype` are preserved.

## Verification

macOS arm64, Node 23.11.0:

- Before fix: `pnpm exec vitest run src/__tests__/searchparams.test.ts` — ten prototype-name regressions fail; 22 existing tests pass.
- After fix, from `packages/farm`: `pnpm exec vitest run src/__tests__/searchparams.test.ts src/__tests__/load-search-params.test.ts src/__tests__/workflows.test.ts` — 56 passed.

Regression coverage checks single/repeated/empty values and JSON round-tripping for ten inherited names. Existing tests also verify prototype safety and use of this helper by the generated production and hydration paths.

## Documentation and release

Query-and-params guide documents raw query representation and reserved keys. No version changes.

---
## Summary by cubic
Fixes search parameter handling so query keys like `toString` and `hasOwnProperty` are treated as ordinary data instead of inherited JavaScript methods. Previously these keys became arrays containing functions and serialized to `null`; now they remain strings or ordered arrays and JSON round-trip correctly.

- The search-params helper only reuses an existing value when the result object owns the property.
- Existing behavior for repeated values, empty strings, insertion order, URL decoding, and blocked prototype keys is unchanged.
- Adds regression coverage and documents that prototype-named keys are ordinary query data.

<sup>Written for commit 71c9a3a234c41f0baf6c336121f7f4f8c2667ed3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/980?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:

PR #992, opened 2026-09-13
feat(wasm): support browser WebAssembly imports and workers

Kinfe123:
## Problem

Apps using prebuilt Wasm ESM modules need more than a single Vite loader. Browser targets, worker builds, async initialization, and production asset URLs all need to agree. Farm did not have an official plugin or an executable example covering that setup.

## Root cause

A plain upstream loader leaves its development helper at a URL Farm treats as a page. Farm also constructs separate production Vite configurations, so worker configuration needs to be carried into that build explicitly. Native top-level await requires a newer browser target than Farm's general default.

## Change

Add `@farm.js/wasm` with a small API:

```ts
import { wasm } from "@farm.js/wasm";

plugins: [wasm()]
// Optional: wasm({ target: ["chrome120", "firefox120", "safari17"] })
```

- Reuse `vite-plugin-wasm` for binary imports and its dependency optimizer support.
- Register the loader helper as a proper Vite virtual module.
- Install fresh loader instances for workers, preserve app worker plugins, and use ES-module workers.
- Use the explicit plugin target, then the app's `vite.build.target`, then the documented native-module baseline. Leave server targets unchanged.
- Add a real WABT-generated demo, precise binary declarations, handled loading/trap errors, worker cleanup, and browser coverage for both Farm builders.

## Safety and compatibility

Opt-in, renderer-neutral browser/worker functionality. No core runtime changes, automatic worker pool, Rust installation, security-header relaxation, or universal server/edge/WASI claim. Invalid options, duplicate registration, and IIFE workers fail clearly. Vite's explicit `?url` and `?init` imports remain supported.

V1 deliberately uses native top-level await. The current upstream compatibility transformer fails with its resolved SWC version ([upstream issue](https://github.com/Menci/vite-plugin-top-level-await/issues/79)); this package does not introduce that dependency or require a transitive pin. Browser feature requirements and older Safari limitations are documented. Browser-only imports remain the app's responsibility, including during SSR.

## Verification

Local environment: macOS arm64, Node 23.11.0, pnpm 8.12.1, Playwright Chromium.

- `pnpm --filter @farm.js/wasm test`: 20 passing tests, including real dev transforms/builds, worker plugin composition, invalid binaries, target errors, and a basic SSR transform smoke test.
- Confirmed the virtual-helper regression fails when its null-prefixed identity is removed and passes with the fix restored. Raw Wasm imports without the plugin are also a negative control.
- `pnpm --filter @farm.js/wasm type-check`, `pnpm --filter @farm.js/wasm build`, and `pnpm --dir examples/wasm-demo type-check`: pass.
- `pnpm test:e2e:wasm:run`: 8 passing dev/production checks with Rolldown.
- `FARM_VITE_BUILDER=rollup pnpm test:e2e:wasm:run`: the same 8 checks pass with Rollup.
- Browser checks cover deferred binary loading, page/worker results, repeated runs, traps and recovery, missing binaries, terminated workers, narrow layouts, and root-relative assets from a base-path page.
- `pnpm docs:build` (Vercel) and `pnpm --dir docs exec farm build --preset node-server`: pass. Production docs and sidebar inspected in Chromium.
- `pnpm --dir docs exec farm generate --check`, the equivalent example check, and `pnpm docs:check-navigation`: pass.
- Focused Oxfmt/Oxlint, browser-test TypeScript check, public-package discovery/packing, and importing the built package export: pass.

CI adds dedicated Chromium jobs for both builders on Node 22.13. The package also participates in the existing package build/test/type-check jobs. Historical browser versions and non-Chromium engines were not executed locally.

## Documentation and release

Added the WebAssembly guide, plugin overview/index/sidebar entries, generated docs routes, package README, and a maintained demo with generated route declarations. The new independently versioned package starts at `0.1.0-beta.0`, matching the existing plugin pattern. Existing workspace build, package-test, example-build, pack, and publish discovery include it without changing the shared release group. No release or existing package version bump in this PR.

### Screenshots

Wasm demo

WebAssembly docs and sidebar

---
## Summary by cubic
Adds `@farm.js/wasm`, an official Farm plugin for importing prebuilt WebAssembly modules in browser code and browser workers. Previously Farm had no supported path for Wasm ESM imports; now apps can use `wasm()` in `farm.config.ts` and get working dev and production builds with both the Rolldown and Rollup Vite builders.

**Details**
- Supports native top-level await and module workers; uses `vite-plugin-wasm` behind the scenes.
- Installs fresh loader instances per worker build, preserves existing worker plugins, and requires `worker.format: "es"`.
- Accepts an optional `target` override; otherwise uses `vite.build.target` or a modern baseline, leaving server targets unchanged.
- Rejects invalid options, duplicate registration, and IIFE workers with clear errors; Vite's `?url` and `?init` imports keep working.
- Adds a runnable example (`examples/wasm-demo`), a WebAssembly guide, and E2E tests for both builders in CI.

<sup>Written for commit 2ec74bceec0ba7caf7daaf2154dfee53e12fbbc5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/992?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Fixed the CI startup timeout. The dev server binds to localhost, which resolves to ::1 on Linux, but the test was probing 127.0.0.1. I reproduced the mismatch locally: the IPv4 probe failed while localhost returned 200. The test now uses the same hostname as Vite, and dev startup output is visible in CI. Both 8-test browser suites pass with NODE_OPTIONS=--dns-result-order=ipv6first. No timeout increase or application behavior change.

PR #1019, opened 2026-09-14
fix(markdown): preserve hydration with server-function examples

Kinfe123:
## Problem

Docs development pages containing `createServerQuery` or `createServerFn` examples return HTTP 500 for their Markdown module requests. The shared docs client cannot load, leaving server-rendered content visible but controls such as the mobile sidebar unhydrated.

This was found while verifying #1016 and is intentionally a separate fix.

## Root cause

The client server-function boundary check runs before the docs Markdown compiler. Its import/call detection scans raw fenced and inline examples as though they were executable application code. Existing boundary tests only covered JavaScript/TypeScript inputs.

## Change

- For `.md` and `.mdx` modules (including query suffixes), parse only when the existing detector finds a potential violation.
- Reuse the existing MDX parser to mask fenced and inline code nodes, then run the same boundary check on the remaining source.
- Add real Vite-transform regression coverage, including five previously blocked documentation pages.
- Document the distinction between displayed examples and executable MDX.

## Safety and compatibility

- Executable MDX declarations and expressions are still rejected without the server-function transform.
- Parse failures retain the original boundary error; Markdown is not blanket-exempted.
- Ordinary JS/TS, SSR handling, and the existing server-actions opt-in are unchanged.
- No new dependency, public configuration, renderer requirement, or runtime/browser code is added. The additional parser work is limited to candidate Markdown violations in the build/dev transform.

## Verification

macOS arm64, Node 24.21.0, pnpm 8.12.1:

- Before the fix, the initial regression suite failed 7 of 10 cases on displayed Markdown examples.
- `pnpm --filter @farm.js/core exec vitest run src/__tests__/markdown-server-query-boundary.test.ts src/__tests__/server-query-boundary.test.ts src/__tests__/app-markdown.test.ts src/__tests__/vite-config.test.ts` — 39 tests passed across 4 files.
- Focused `pnpm exec oxfmt --check` and `pnpm exec oxlint` on changed files — passed; `git diff --check` — passed.
- `pnpm docs:check-navigation` — passed.
- `pnpm docs:build` — passed, including core/CLI runtime builds, Prisma client generation using a dummy localhost URL (no database connection or migration), and the Vercel production output.
- Core ESM/CJS builds passed. Its declaration worker exceeded the default heap; `NODE_OPTIONS=--max-old-space-size=8192 pnpm --filter @farm.js/core exec tsup --dts-only` then passed.

Reproducible browser check with the rebuilt framework: start the docs dev server without enabling server actions, open `/docs/server-queries`, and verify the Markdown modules for `cache-ppr`, `plugins/content`, `renderers`, `routing`, and `server-queries` return 200 instead of 500. At 390×844, open the sidebar, scroll to and click **Cache and PPR**, confirm `/docs/cache-ppr` and its heading, then open Search and confirm the search combobox appears. These steps passed with no captured browser errors or Vite overlay.

The isolated dev server still logs existing extensionless telemetry-source resolution warnings during route discovery; they do not prevent the tested docs client from loading. Unrelated warnings were not changed.

## Documentation and release

Updated the Server Queries guide. No generated route changes, template changes, version bump, or release changes.

---
## Summary by cubic
Fixes docs Markdown pages returning HTTP 500 when they contain server-function examples, so docs client hydration no longer breaks.

Markdown/MDX fenced and inline code examples are now treated as displayed content and skipped by the server-function boundary check. Executable MDX declarations and expressions are still rejected, and invalid Markdown never suppresses a real violation. Added regression tests covering multiple fence forms and five previously blocked docs pages.

<sup>Written for commit d27f6d613ea5adb7c1faf6091118f2e91d7bd80c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
closing in favor of #1062, which landed a simpler fix for the same false positive: markdown ids are now exempt from the client boundary guard entirely, so fenced examples can't trip it regardless of what they show. the extra note this added to the server-queries docs page might still be worth a tiny follow-up if we want it.

PR #1070, opened 2026-09-15
fix(auth): stop signing integration sessions with a published dev secret

Kinfe123:
## Problem

`@farm.js/auth0` and `@farm.js/workos` sign their session cookies with a **hardcoded secret published in this repository** whenever no explicit secret is configured and `NODE_ENV !== "production"`:

```ts
// farm-auth0
const DEV_SECRET = "farmjs-auth0-development-secret-2026";
const secret = input.secret ?? process.env.AUTH0_SECRET
  ?? (process.env.NODE_ENV === "production" ? "" : DEV_SECRET);

// farm-workos
const DEV_COOKIE_PASSWORD = "farmjs-workos-cookie-password-development-2026";
```

Anyone can read these constants from the open-source package and forge a validly-signed session cookie for any deployment that runs without `NODE_ENV === "production"` and without a configured secret — staging and preview environments, serverless/edge runtimes that don't set `NODE_ENV`, or a production box where the variable is simply unset. The production hard-throw only fires when `NODE_ENV` is exactly `"production"`, so a misconfigured production silently falls back to the public secret instead of failing.

## Root cause

The dev fallback is a fixed value shipped in the source. Because it is identical across every install, a session signed with it in one place validates everywhere, which is the definition of a forgeable credential. Gating solely on `NODE_ENV === "production"` also misses the common case where that variable is absent at runtime.

## Change

Replace the published constant with a **per-process random fallback**, in a shared, tested core helper so both integrations behave identically:

- Added `resolveIntegrationSessionSecret` to `@farm.js/core` (`integrations.ts`, alongside `defineIntegration`). Precedence: explicit `configured` value, then the first non-empty env var. When neither is set, production throws (a real secret is required), and development uses `randomBytes(32)` memoized per process — never a value in the repository. The fallback is stable for the process lifetime (so signing and verifying agree) but unique to it, and it warns once, telling the developer to set the env var for multi-instance or restart-sensitive deployments.
- `farm-auth0` and `farm-workos` now call the helper (and the `DEV_SECRET`/`DEV_COOKIE_PASSWORD` constants are gone). The helper is resolved after each integration's existing domain/clientId checks, preserving current error precedence and messages.

Zero-config local development still works (the fallback is generated automatically); it simply can no longer be forged from a known constant, and exposed multi-instance deployments fail safe (sessions do not validate across instances until an explicit secret is set) rather than silently accepting forgeable cookies.

The helper lives in `@farm.js/core` rather than `@farm.js/integration-utils` because both integration packages already depend on core, and core carries the test infrastructure that makes the security contract verifiable; `integration-utils` currently ships without a test runner.

## Safety and compatibility

Server-only integration configuration. No public API removal; one added core export (`resolveIntegrationSessionSecret`). Behavior is unchanged when a secret is configured (explicit input or env var) and in production (still throws when none is set). The only behavior change is the development-without-a-secret path, which is now safe. Affects `@farm.js/auth0` and `@farm.js/workos`.

## Verification

- New test `packages/farm/src/__tests__/integration-session-secret.test.ts` (6 cases): explicit-over-env precedence, first-non-empty env, production throws when unset, the dev fallback is **not** either published constant and is a 64-char hex value, stable per integration but unique across integrations, and warns once. Passes.
- `pnpm --filter @farm.js/core test` (this suite) and `pnpm --filter @farm.js/core type-check` — clean. oxfmt and oxlint clean on changed files (the two `no-unused-vars` warnings in `integrations.ts` are pre-existing dead code — `matchesMatcher`/`matchesPath` — outside this diff).
- No dependency or lockfile changes.

## Documentation and release

None. No documented API relied on the removed constants.

---
## Summary by cubic
Stops `@farm.js/auth0` and `@farm.js/workos` from signing session cookies with a hardcoded dev secret published in the repo. With no secret configured, both now fall back to a per-process random value, so cookies can no longer be forged with a known constant; explicitly configured secrets, environment variables, and the production error when neither exists are unchanged.

**Notes**

- The random fallback is stable only within one process, so `AUTH0_SECRET` / `WORKOS_COOKIE_PASSWORD` should be set for multi-instance or restart-sensitive deployments.
- Both integrations now resolve the secret via `resolveIntegrationSessionSecret` in `@farm.js/core`, which warns once when using the fallback.
- Adds tests for precedence, production errors, fallback uniqueness, and warning behavior.

<sup>Written for commit 96272c1c65b7c5a2980141e32ff2e0b6874391ad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Superseded by #1063, which fixed the same vulnerability (dev secret used when NODE_ENV is unset) by failing closed unless NODE_ENV is explicitly development/test. Closing this competing approach since the core issue is already resolved on main.

PR #1102, opened 2026-09-16
fix(image): coalesce identical concurrent cache misses

Kinfe123:
> Stacked on #1101 → #1091 (targets `fix/image-cache-key-negotiated-format`), which touch the same handler. Retarget to `main` once those merge.

## Problem

Identical concurrent cache misses each run the full pipeline. Two simultaneous requests for the same uncached image fetch the origin twice and run the codec twice; a burst on a cold cache (a page going live, a CDN cold start, a crawler) multiplies origin traffic and CPU by the number of in-flight requests.

## Root cause

The handler checks the memory cache and, on a miss, immediately does the work. Nothing records that an identical optimization is already running, so every concurrent miss duplicates it. The cache is only populated after the transform completes, which is exactly the window the duplicates occupy.

## Change

Added a handler-scoped in-flight map keyed by the same cache key. The first miss starts the work; identical concurrent misses await that promise instead of starting their own. The entry is removed when the work settles, so later requests either hit the populated cache or start fresh work.

Abort handling is the subtle part and is handled explicitly:

- the shared work runs under its **own** `AbortController`, not the first request's signal, so one caller disconnecting cannot cancel the image other waiters are still expecting;
- each waiter still races its own request signal, so an aborted request stops waiting immediately rather than being held to the end of the shared work;
- waiters are ref-counted and the shared controller is aborted only when the last waiter leaves, so a fully abandoned burst still stops promptly instead of running to completion.

## Safety and compatibility

No API, response, or configuration change. Sequential requests behave exactly as before. Errors propagate to every waiter of that key, matching what each would have produced on its own, and a failed optimization is not cached (the in-flight entry is simply removed).

## Verification

Added tests to `image-server.test.ts`:
- two overlapping requests for the same uncached image produce one transform and one origin fetch, with both responses `200` (fails without the fix: two of each);
- when one of two overlapping requests is aborted mid-transform, the other still receives a `200` — pinning the guarantee that shared work is not tied to a single caller's signal.

## Documentation and release

None. Internal request handling.

---
## Summary by cubic
Coalesces identical concurrent image cache misses so a burst of requests for the same uncached image performs one origin fetch and one transform instead of one per request.

- Shared work runs under its own `AbortController`, so one caller disconnecting can't cancel the image other waiters are still expecting.
- Each waiter still stops on its own abort, and the shared work is canceled only after the last waiter leaves.
- The shared promise stays marked as handled so a fully abandoned burst can't trigger an unhandled rejection.
- Errors propagate to every waiter, and failed optimizations are not cached.
- Sequential requests are unaffected.
- Adds tests covering one fetch/transform for overlapping requests and serving remaining waiters after an abort.

<sup>Written for commit b0fbd0dc6e7eb354105a52ec7ade094cfb475c9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Pushed b0fbd0dc to fix the Test matrix failure here.

All 33 assertions were already passing; the run failed on an unhandled rejection:

```
Unhandled Rejection
TypeError: Cannot read properties of undefined (reading 'status')
 ❯ fetchImageSource src/image-server.ts:405:54
Tests 2278 passed | 1 skipped — Errors 1 error
```

Mechanism: every waiter can detach before the shared optimization settles. `raceRequestAbort` returns early for an already-aborted signal without ever attaching to the shared promise, and the last waiter leaving calls `controller.abort()`, so `run()` rejects with no listener.

Fix is one no-op handler on the shared promise. I checked it does not swallow real failures: with a transform that throws, both coalesced waiters still get a 5xx, so the rejection reaches every caller.

Reproduced locally before the fix (33 passed, 1 error) and after (33 passed, no error). oxlint reports one pre-existing warning in this file, unrelated to the change.

PR #1103, opened 2026-09-16
fix(telemetry): enforce body limits while reading, not after

Kinfe123:
## Problem

The public telemetry endpoints buffer the entire request body before enforcing their size limit. `await request.text()` runs to completion first, and only then is the 8 KB ceiling checked, so an unauthenticated caller can make the docs server hold an arbitrarily large body in memory. The `Content-Length` pre-check does not close this: the header is absent for chunked uploads and can simply understate the real body.

The same pattern appears in all three endpoints: `v1/sites`, `v1/events`, and `dashboard/session`.

## Root cause

The limit was applied to an already-materialized string (`new TextEncoder().encode(text).byteLength > MAX_BODY_BYTES`). By the time that check can run, the memory has already been allocated — the check reports the abuse instead of preventing it.

## Change

Added `readTextWithLimit` in `docs/src/lib/request-body.ts`, which keeps the `Content-Length` fast path and then reads the body through its stream, aborting as soon as the accumulated size passes the ceiling and cancelling the reader. All three endpoints now use it, so the limit is enforced as bytes arrive rather than after they have all been accepted.

Response codes and bodies are unchanged (`413` with the same payload per endpoint).

## Safety and compatibility

Behavior for well-formed requests is identical — same limits, same status codes, same parsed values. Only the point at which an oversized body is rejected changes. The helper returns an empty string for a bodyless request, matching `request.text()`.

## Verification

Added a regression test to `v1/sites/route.test.ts`: a chunked body with no `Content-Length` that offers 4 MB is rejected with `413`, and the test asserts the stream was only pulled a handful of times (`< 64` KB-sized chunks) rather than drained. Without the fix the whole 4 MB is buffered before the check and the pull-count assertion fails.

## Documentation and release

None. Internal endpoint hardening in the docs app.

---
## Summary by cubic
Enforces telemetry body size limits while the body is being read instead of after `request.text()` has buffered it, so an unauthenticated caller can no longer make the docs server hold an arbitrarily large body in memory.

- Adds `readTextWithLimit` in `docs/src/lib/request-body.ts` and uses it in `v1/sites`, `v1/events`, and `dashboard/session`.
- Keeps the `Content-Length` fast path and now stops reading chunked bodies, or bodies whose `Content-Length` understates their size, as soon as they pass the limit.
- Leaves valid request behavior and 413 responses unchanged.
- Adds a regression test that proves an oversized chunked body is rejected without being drained.

<sup>Written for commit 532a4cf49d2bddb3d2b36ea886eb5d454cd51f74. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Pushed 532a4cf4 to fix the Framework E2E failure.

The job never got to the browser tests; the docs build could not load a route:

```
❌ Error loading route docs/src/app/api/telemetry/dashboard/session/route.ts:
   Failed to load url ../../../../lib/request-body
```

`dashboard/session/` is five directories below `docs/src`, same as the `v1/events` and `v1/sites` routes that import the same helper with five levels. This one used four, so it resolved to `docs/src/app/api/lib/request-body`, which does not exist.

Verified the five-level specifier resolves to the real `docs/src/lib/request-body.ts`, and the docs telemetry tests pass (7 passed).

PR #1108, opened 2026-09-16
fix(workos): bind the OAuth callback to the browser that started the flow

Kinfe123:
> Stacked on #1106 (which is stacked on #1097). Base is `fix/auth0-workos-logout-csrf` to avoid conflicts in the same file; retarget down the stack as each merges.

## Problem

`@farm.js/workos`'s OAuth callback never validated `state`. It read the value only to recover a redirect target:

```ts
const parsed = JSON.parse(rawState) as { returnTo?: string };
returnTo = getReturnTo(parsed.returnTo ?? null, "/dashboard");
```

Any authorization code that reached the callback was exchanged and sealed into the requesting browser's session cookie. So an attacker could:

1. Start a normal WorkOS sign-in against the app and authenticate as themselves.
2. Stop at the redirect and keep the `code` instead of letting the browser follow it.
3. Send a victim to `https://app.example.com/callback?code=<attacker_code>`.

The victim's browser receives `Set-Cookie: wos-session=<attacker's sealed session>` and is now signed in as the attacker. Anything the victim then writes lands in the attacker's account, and they can read it at will. This is the session-fixation half of login CSRF, the same impact class as #1097, and it needs no cooperation from the victim beyond opening a link.

`@farm.js/auth0` was never affected: it already binds `state` to a signed cookie and carries a `codeVerifier`.

## Root cause

The integration treated `state` purely as a transport for `returnTo`, so it was attacker-supplied data being parsed rather than a value the app had issued and could recognise. There was nothing tying the callback to the browser that began the flow.

No test caught it because every existing callback test drove a well-formed flow; none asserted that a code from a *different* flow is refused.

An origin check cannot fix this. The callback is legitimately a top-level cross-site navigation from `workos.com`, so it looks identical to the attack in origin terms. The fix has to be state binding.

## Change

Mirrors the approach already proven in `@farm.js/auth0`:

- Sign-in and sign-up generate `randomBytes(32)` as `state` and pass it to `getAuthorizationUrl`.
- `{ state, returnTo }` is HMAC-signed with the existing `cookiePassword` and stored in a short-lived cookie (`wos-session-state`, `maxAge` 600).
- The callback unsigns that cookie and requires `statePayload.state === state` before calling `authenticateWithCode`. A missing, mismatched, or unsigned cookie returns `400` before any session exists.
- `returnTo` now comes from the signed payload instead of the query value, so it is no longer tamperable in transit.
- The state cookie is cleared on the successful callback, making it single use.

`Response.redirect()` cannot carry a `Set-Cookie`, so the two redirect handlers now build `new Response(null, { status: 302, headers })`. The JSON client responses carry the same headers.

**PKCE is deliberately not added.** The SDK documents `codeVerifier` on `authenticateWithCode` as *"When provided, client_secret is not sent (public client mode)"*. Farm constructs WorkOS as a confidential client with an `apiKey`, so adding a verifier would change how the token request authenticates and could break working deployments. State binding closes this vulnerability on its own; moving to PKCE is a separate decision about client mode.

## Safety and compatibility

- Sign-in flows already in flight when this deploys hit the new check once and get `400`; retrying signs in normally. This is called out in the docs.
- Apps relying on reading `returnTo` out of the raw `state` query value would need to stop; `state` is now opaque. That value was never a supported API surface.
- No new configuration and no new dependency. The signing secret is the `cookiePassword` the integration already requires.
- Unchanged: session reads, sign-out, protected-route middleware, and the app-owned `instance` path.
- **Known limit:** the signing helpers are copied from `@farm.js/auth0` rather than shared. Two copies now exist; hoisting them into `@farm.js/integration-utils` is worth doing, but refactoring both packages inside a security fix would widen the blast radius, so it is left as a follow-up.

## Verification

macOS, Node v23.11.0, pnpm 8.12.1, `@workos-inc/node` 8.9.0.

```bash
pnpm --filter @farm.js/core exec vitest run src/__tests__/workos-callback-state.test.ts
pnpm --filter @farm.js/workos type-check
pnpm exec oxlint packages/farm-workos/src/index.ts
pnpm docs:check-navigation
```

6 new tests pass; 55 pass across the 5 suites this stack touches.

**Failing without the fix.** Restoring the original callback, keeping the tests, fails 5 of 6:

- issues an unguessable state and stores it in a signed cookie
- rejects a callback with no state cookie, without creating a session
- **rejects an attacker's code replayed into a victim's browser**
- rejects a forged state cookie that was not signed by this app
- completes a genuine callback and clears the state cookie

The third is the vulnerability itself: it runs two independent flows, presents the attacker's `state` alongside the victim's cookie, and asserts `authenticateWithCode` is never called and no `Set-Cookie` is issued.

The sixth test (`returnTo` from the signed cookie) passes either way — the old code reached the same redirect by parsing JSON state — so it guards the behaviour rather than the bug.

## Documentation and release

- `docs/src/app/docs/integrations/auth/workos/page.md`: new "OAuth state binding" section covering the cookie, single use, the opaque `state`, and the one-time failure for in-flight sign-ins during deploy.
- No release-flow, template, example, or generated-artifact changes.

---
## Summary by cubic
Fixes a session-fixation vulnerability in `@farm.js/workos` where a code from an attacker's sign-in flow could be replayed into a victim's browser, signing the victim into the attacker's account. The callback now accepts a code only when the browser presents a matching signed, single-use state cookie.

**Bug Fixes**
- The callback rejects missing, mismatched, or forged state with `400` before calling `authenticateWithCode` or creating a session.
- `returnTo` now comes from the signed state cookie instead of the raw `state` query value, so it can't be tampered with.
- The state cookie is cleared on success and uses the existing `cookiePassword`, with no new dependency.
- PKCE is intentionally not added: sending a `codeVerifier` would stop `@workos-inc/node` from sending `client_secret` and switch this confidential client to public-client mode.

**Migration**
- In-flight sign-ins fail once when this deploys; retrying signs in normally.
- `state` is now opaque; apps that parsed `returnTo` out of the raw `state` query value must stop.

<sup>Written for commit 93fb4bbddec3be6b6cbd3d9e22051eed068ba52a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Superseded by #1087, which landed the same fix while this was in review. Main already binds the callback to a random signed state cookie, verifies it before authenticateWithCode, takes returnTo from the signed payload, and clears the cookie on success. Its unsignState also type-checks the decoded fields, which this version did not. Closing rather than pushing a competing implementation.

PR #1109, opened 2026-09-16
refactor(integration-utils): share the auth cookie signing helpers

Kinfe123:
> Top of the stack: #1097 → #1106 → #1108 → this. Base is `fix/workos-callback-state-binding`; retarget down the stack as each merges. This is the follow-up #1108 called out as a known limit.

## Problem

`@farm.js/auth0` and `@farm.js/workos` each carried their own copy of the HMAC sign/verify pair used for cookie payloads. The copies were byte-identical — I introduced the second one in #1108 by copying auth0's.

That code decides whether an attacker-reachable cookie is trustworthy: base64url encoding, HMAC comparison, the length guard that stops `timingSafeEqual` from throwing, and the JSON parse that must not leak a partial value. Duplicated security code drifts. A fix or hardening applied to one copy silently leaves the other behind, and nothing in the build or tests would flag the divergence.

No behavior is wrong today. This removes the conditions for a future bug rather than fixing a current one.

## Change

`signCookieValue` and `unsignCookieValue` move into `@farm.js/integration-utils`, which AGENTS.md designates as the home for provider-neutral integration helpers. They live in `cookies.ts` next to `createRequestCookie`/`getCookieValue`, since that is what they exist to serve.

Both packages import them and their local copies are deleted. The shared implementation is the same code both packages already ran, so this is a move, not a rewrite:

- auth0: session payloads and the OAuth state cookie
- workos: the OAuth state cookie added in #1108

`unsignCookieValue` additionally accepts `null` (not just `undefined`), which removes the `?? undefined` workaround #1108 needed at the `getCookieValue` call site.

Now-unused `createHmac` and `timingSafeEqual` imports were dropped from both packages. auth0 keeps `createHash`/`randomBytes` for PKCE, workos keeps `randomBytes` for state.

The doc comment states plainly that this provides integrity and not confidentiality — the payload is encoded, not encrypted, and is readable by anyone holding the cookie — so nobody later signs a secret expecting it to be hidden.

## Safety and compatibility

- No behavior change. Same algorithm, same encoding, same failure modes, so cookies issued before this change still verify.
- Public API is additive: two new exports from `@farm.js/integration-utils`. Nothing is removed; the deleted functions were module-private.
- No new dependency. `node:crypto` moves from the two packages into the shared one.
- No effect on renderers, runtimes, or deployment targets.

## Verification

macOS, Node v23.11.0, pnpm 8.12.1.

```bash
pnpm --filter @farm.js/core exec vitest run src/__tests__/integration-cookie-signing.test.ts
pnpm --filter @farm.js/integration-utils type-check
pnpm --filter @farm.js/auth0 type-check
pnpm --filter @farm.js/workos type-check
pnpm --filter @farm.js/auth0 build
pnpm --filter @farm.js/workos build
pnpm exec oxlint packages/farm-auth0/src/index.ts packages/farm-workos/src/index.ts packages/farm-integration-utils/src/cookies.ts
```

The refactor's real test is that the callers are unchanged in behavior. All suites that exercise these paths pass: **61 tests across 6 files**, covering the workos callback state binding, auth0 and workos sign-out, supabase auth routes, integration instance injection, and server-action security. The pre-existing `auth0-secret-resolution` and `workos-cookie-password-resolution` suites also pass (18 tests with the workos callback suite).

6 new tests cover the helper directly, including the cases that matter for a security primitive and were previously untested in either package:

- round-trips a payload
- rejects a payload edited in transit (swapped body, original signature)
- rejects a value signed with a different secret
- rejects missing, empty, and unsigned values
- rejects a signature of the wrong length **without throwing** — guards the `timingSafeEqual` length precondition
- rejects a correctly signed payload that is not valid JSON

I also confirmed the built output actually resolves to the shared helper rather than an inlined leftover: `unsignCookieValue` appears in both `packages/farm-auth0/dist/index.js` and `packages/farm-workos/dist/index.js`.

## Documentation and release

None. No user-facing API or behavior changes; the new exports are for integration authors and are documented by their doc comments.

---
## Summary by cubic
Moves the duplicated HMAC cookie signing helpers out of `@farm.js/auth0` and `@farm.js/workos` into `@farm.js/integration-utils` so the security-sensitive code stops existing in two copies. Behavior is unchanged — the shared implementation is the same code both packages already ran — so cookies issued before this change still verify and no migration is needed.

**Refactors**
- Both packages now import `signCookieValue` and `unsignCookieValue` from `@farm.js/integration-utils`; their local copies and unused `node:crypto` imports are deleted.
- `unsignCookieValue` also accepts `null`, removing the `?? undefined` workaround in the WorkOS callback handler.
- The changes are additive: two new exports, no removed public API, no new dependencies.
- A doc comment now states the signing provides integrity, not confidentiality.
- Adds 6 tests covering the shared helper directly, including the `timingSafeEqual` length guard and rejection of tampered values.

<sup>Written for commit d1a2528912b089590fff03dc111b47e95b3995b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Closing and rebuilding. This branch was stacked on #1108, which is now closed as superseded by #1087, so its diff no longer applies.

The underlying point still stands, and is now sharper: main has two different implementations of the same HMAC cookie signing. farm-auth0 has the generic signValue/unsignValue, and #1087 added workos's signState/unsignState, which additionally type-checks the decoded fields. Consolidating them is still worth doing, but it is a real refactor against main rather than the straight move this branch contained. Will reopen against main once the sign-out guard in #1106 lands, to avoid two PRs editing the same regions.

PR #1113, opened 2026-09-16
fix(image): call the handler factory correctly in the cache key test

Kinfe123:
## Problem

`Test (ubuntu-latest, Node 22.13)`, `Test (ubuntu-latest, Node 24)`, and `Test (windows-latest, Node 24)` have been failing on `main` since #1101:

```
FAIL src/__tests__/image-server.test.ts > image cache keys >
     shares one cache entry across Accept headers that negotiate the same format
TypeError: Cannot read properties of undefined (reading 'fetch')
 ❯ createFarmImageHandler src/image-server.ts:78:27
```

I hit this on two unrelated PRs (#1106 and #1110). #1110 is test-only and changes no source, which is what identified it as pre-existing rather than caused by either branch.

## Root cause

`createFarmImageHandler` takes two positional arguments:

```ts
export function createFarmImageHandler(
  config: ResolvedFarmImageConfig,
  options: CreateFarmImageHandlerOptions,
): FarmImageHandler
```

The test added in #1101 passed a single merged object instead, with `config` as a property:

```ts
const handler = createFarmImageHandler({
  config: resolveFarmImageConfig({ ... }),
  transform,
  fetch: fetcher as typeof fetch,
  fetchRemote: fetcher as typeof fetch,
});
```

So `options` was `undefined` and the factory threw on `options.fetch` at line 78, before the test reached a single assertion. Every other call in the same file uses the two-argument form correctly.

The call type-checks because `ResolvedFarmImageConfig` is structurally satisfied loosely enough at that position, so `tsc` did not catch it and the failure only appears at runtime. **The test has never passed**, which also means #1101's fix shipped without working verification.

## Change

Split the argument into `(config, options)`. No source change, no change to what the test asserts.

## Verification

macOS, Node v23.11.0, pnpm 8.12.1.

```bash
pnpm --filter @farm.js/core exec vitest run src/__tests__/image-server.test.ts
```

31 pass, up from 30 passing and 1 erroring.

**The repaired test now actually tests #1101.** With the source change from that PR reverted (`git show be205f29^:packages/farm/src/image-server.ts`), it fails on the assertion rather than the constructor:

```
× shares one cache entry across Accept headers that negotiate the same format
  → expected 4 to be 1
```

Four transforms instead of one is exactly the cache-key bug #1101 fixed: each `Accept` variant negotiating to `image/webp` was getting its own cache entry. Restoring the source change makes it pass. So the fix is sound; only its test was wrong.

## Documentation and release

None. Test-only.

---
## Summary by cubic
Fixes the image cache key test so it actually verifies the cache-key fix instead of failing before any assertion runs. The test called `createFarmImageHandler` with one merged object, leaving `options` undefined and making the factory throw; it now passes the two expected arguments `(config, options)`.

- Test-only change; no source behavior changes.
- With the original fix reverted, the test now fails on the assertion (4 transforms instead of 1), confirming it exercises the intended behavior.

<sup>Written for commit 750dd3fd774d0eb25d3930e69582c20dcec5badf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Superseded by #1112, which fixed the same call shape while this was open. Main is green on image-server again.

PR #1118, opened 2026-09-17
perf(react): avoid temporary keyed-row entry arrays

Kinfe123:
## Problem

Rebuilding the compiler's keyed-row lookup creates a temporary array of `[key, row]` pairs. A 10,000-row rebuild therefore creates 10,000 pair arrays plus an outer staging array, in addition to the Map it actually needs.

This is a separate allocation-reduction follow-up, not the batch-insert change in draft #1116, and it does not establish the cause of that PR's queued-resize performance-gate miss.

## Root cause

The private `keyedRowInstancesByKey` helper used `new Map(instances.map(...))`. Existing behavior tests checked the resulting rows, but not these intermediate allocations.

## Change

Build a fresh native Map with forward `Map.set` calls over the same framework-owned row instances. No caller or eligibility condition changes. Four new regression tests cover three consecutive single/queued resized-window commits in static and hybrid modes. Four source-based compiler tests also verify that the actual optimized queued-resize benchmark emits two window hints and the snapshot control emits none.

## Safety and compatibility

- Preserve insertion order, key/instance association, fresh-map ownership, index maintenance, cleanup, preparation-before-commit and existing fallback decisions.
- Rebuilds remain O(n); this is not a constant-time map update or a claim of fewer DOM changes.
- No application-owned array method is bypassed. The helper receives framework-owned row-instance arrays.
- No public API, option, dependency, compiler eligibility, renderer-neutral path or benchmark threshold changes.
- Compiler-off runtime-size fixtures are unchanged. Existing hydration, unmount, native-error, identity, focus/selection, fallback and randomized differential coverage remains enabled.

## Verification

Environment: Darwin 25.2.0 arm64 / Apple M1; Node 23.11.0 for browser/unit/build work; Chromium 145.0.7632.6. Runtime-size checks use Node 24.19.0, zlib 1.3.2.1-motley-3246f1b, for both parent and candidate. Tested base: main `4753f3a1ab1717a2b8940e485b2861a0060d71db`; candidate `eab816d867b2c306198b56329d81b3fcb4812851`.

- Regression-first: all four new allocation tests fail on the unchanged helper (256 or 257 temporary row-entry arrays), after row values/order, retained DOM identity, removal and callback assertions pass. The same tests pass with the loop.
- `pnpm --filter @farm.js/react exec vitest run --maxWorkers=2`: 967 passed, 14 existing skips.
- `pnpm --filter @farm.js/react test:stress`: 23 passed.
- `pnpm --filter @farm.js/react type-check`: passed.
- `pnpm --filter @farm.js/react test:compat`: packaged React 18.3.1 and 19.2.8 passed.
- Build, focused oxfmt/oxlint, docs generation/navigation checks and `pnpm docs:build`: passed.
- Unchanged `packages/farm-react/scripts/benchmark-runtime-size.mjs --check` under the same Node 24 binary: passed for parent and candidate. Window gzip 74,573 → 74,572 bytes; structural append premium 17,538 → 17,538 against the unchanged 17,541 limit. Shared fixtures gain 21 raw bytes, gzip changes range −1 to +6 bytes, some Brotli fixtures increase. No blanket bundle-size reduction claim.

Matched production A/B: fresh browser per trial, four counterbalanced rounds, five warmups and ten measured samples per action. Same production builds, source workload and DOM correctness assertions; no overlapping builds/unit suites. All correctness checks passed. Retain both reports, not just the favorable samples:

| Comparison | Static hinted median | Hybrid hinted median | Static snapshot median | Hybrid snapshot median |
| --- | --- | --- | --- | --- |
| Initial | 9.7 → 11.0 ms | 10.1 → 9.5 ms | 21.2 → 23.5 ms | 22.8 → 21.2 ms |
| Predeclared confirmation | 9.7 → 9.8 ms | 9.6 → 9.5 ms | 21.8 → 22.4 ms | 21.8 → 21.9 ms |

The results do not establish a reliable latency win or universal non-regression. The shared helper also affects the snapshot control; that control is not independent evidence of machine noise. No further repetitions are planned to select a favorable result.

Full unchanged dashboard benchmark: correctness PASS, 46/47 performance gates PASS, exit 1. Hybrid reversal misses the unchanged snapshot floor: 32.1 ms versus snapshot 39.0 ms gives 1.214953x, below 1.25x. Its React speedup is 11.281931x (8x minimum). Static reversal passes. Queued resize passes in static/hybrid (5.283186x/4.627907x React, 2.017699x/1.821705x snapshot). All compiled owner execution counts remain zero and both 10k/20k update persistence gates exceed the unchanged 8x floors. Default samples, controls and all 47 performance gates are preserved.

Full benchmark command (with the pinned Chromium 145 executable supplied through the environment):

```sh
FARM_EXPERIMENT_BROWSER_PATH="<Chromium 145 executable>" \
NODE_OPTIONS=--max-old-space-size=8192 \
FARM_DASHBOARD_REPORT=/tmp/react-keyed-map-benchmark.json \
pnpm --filter farm-react-compiler-dashboard-example benchmark
```

**Draft; not merge-ready.** Published for review at the maintainer's request. The full-run miss does not establish that this helper caused a regression, but it prevents claiming that all required performance floors are preserved. Do not retry until green or weaken the floor; a separate matched reversal diagnostic is the planned follow-up. The earlier mixed A/B evidence remains part of the record.

## Documentation and release

Updated the experimental React compiler documentation and the existing compiler-dashboard example README to explain the allocation reduction and unchanged linear work. No new example, generated artifact, version, tag, starter or release-flow changes.

---
## Summary by cubic
Replaces the temporary `[key, row]` pair arrays in the keyed-row map rebuild with direct `Map.set` calls, reducing allocations on large row rebuilds. The old helper built `new Map(instances.map(...))`, creating one pair array per row plus a staging array; the new one fills a fresh Map in a forward loop, so insertion order, key/instance association, and O(n) work are unchanged.

- Adds four regression tests covering repeated single and queued resized windows in static and hybrid modes.
- Adds compiler control tests verifying the optimized queued-resize handler emits two window hints and the snapshot control emits none.
- Updates React compiler docs and dashboard README to document the allocation reduction.

Benchmark results are mixed; the hybrid reversal misses the unchanged snapshot performance floor, so this is not a validated latency improvement and remains a draft.

<sup>Written for commit eab816d867b2c306198b56329d81b3fcb4812851. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Follow-up diagnostic on the hybrid reversal miss (no runtime changes):

Compared the frozen old-helper/new-helper production builds with identical source workload and browser, four counterbalanced fresh-browser rounds, five warmups and ten measured samples per action. Both modes use the real optimized reversal and snapshot-control buttons. Snapshot source hashes match main `4753f3a1` and this PR's `eab816d8`.

| Mode | Old → new hinted median | Old → new snapshot median |
| --- | --- | --- |
| Static | 23.9 → 22.8 ms | 36.0 → 36.6 ms |
| Hybrid | 23.9 → 23.4 ms | 37.4 → 35.9 ms |

All 480 reversals preserved every row's DOM identity and the expected order/content, with zero extra compiled owner executions or browser errors. The timed boundary retains the full benchmark's endpoint/revision predicate; stronger complete-row checks run outside timing.

This **does not reproduce a consistent hybrid slowdown**: the new helper was faster in two rounds and slower in two. It also **does not clear the failed full-suite gate**, because this diagnostic does not reproduce the suite's entire preceding workload history. The original 46/47 report and mixed resize comparisons remain valid evidence. No threshold was changed and no repeated run was used to seek a green result.

An independent local follow-up now adds nine source-derived tests that require exactly one reorder hint for the actual optimized handler and zero for its control in each mode, plus native-handler equivalence across three repeated reversals of frozen arrays from zero through 10,000 rows. Focused compiler/control/runtime coverage: 40 passed, one existing stress-only skip; type-check and lint passed. These are protective tests, not a runtime fix. **This PR remains a draft and is not merge-ready.**

Kinfe123:
The requested matched before/after investigation is complete. **Keep this PR draft: equal-or-better performance has not been established.** No gates, controls, sample counts, or benchmark source were changed.

Replayed the original production benchmark callback verbatim through reversal, retaining all 46 preceding/current measurements and correctness assertions: four counterbalanced rounds, 16 fresh Chromium 145 browser trials, frozen parent `4753f3a1` and candidate `eab816d8` builds, both static/hybrid modes, 60 dashboard samples × 10 updates, 10 table samples, five warmups. Node 23.11.0, Apple M1, Darwin 25.2.0 arm64. All browser correctness checks passed, with zero added compiled owner executions or browser errors.

Pooled before → after medians:

| Operation | Static | Hybrid |
| --- | --- | --- |
| Native reversal | 24.8 → 22.6 ms | 27.4 → 25.1 ms |
| Reversal snapshot control | 39.0 → 35.3 ms | 40.7 → 38.9 ms |
| Rolling window | 19.2 → 21.3 ms | 19.1 → 19.6 ms |
| Mapped rolling window | 20.0 → 21.6 ms | 18.9 → 21.6 ms |
| Mapped rolling chain | 6.2 → 6.5 ms | 5.8 → 6.3 ms |
| Queued window resize | 14.5 → 15.0 ms | 14.9 → 15.2 ms |

Reversal was faster in three of four rounds per mode. However, hybrid mapped chains were slower in all four rounds, and hybrid mapped rolling windows were slower in three. Many other operations had mixed directions. This is not evidence of an everywhere improvement, nor proof of the engine-level cause. Both the optimized path and snapshot control use the changed helper.

The original full result remains **46/47 performance gates**, with the hybrid reversal snapshot-ratio miss unresolved. The prefix diagnostic cannot replace that full result. All raw samples and all 46 metrics are retained locally in `react-keyed-map-history-ab-2026-09-17.json`, alongside the original full and isolated reports.

A separate, narrower follow-up will investigate adopting the already-owned map returned by validated reorder preparation instead of cloning it again. It starts from main without this shared-helper change and must independently pass correctness, size, and performance checks.

Kinfe123:
rebased onto current main (the conflict was additive test families in compiler-dashboard-controls, both kept), but holding the merge: the full dashboard benchmark on the rebased branch reproduces the queued-resize miss this pr's description already flagged in september. today's run: CORRECTNESS_PASS_PERFORMANCE_REGRESSION with keyedQueuedWindowResizeHintGate FAIL (needs 4x / 1.5x snapshot), while three other runs on the same machine today (main twice, #1382) all pass that gate. so the forward set loop really does depress the queued-resize path, likely because v8 does better with the pair-array Map constructor there than the win from skipping the temporary arrays. unit suites, type-check, and the size gate are all green, so the branch is current and ready whenever the mechanism is understood.

PR #1136, opened 2026-09-17
fix(cron): constant-time secret comparison

Kinfe123:
## Problem

The cron and workflow request-auth gates compared the configured secret with `===`:

- `packages/farm/src/cron.ts` (`isCronRequestAuthorized`): `bearer === secret || headerSecret === secret`
- `packages/farm/src/workflows.ts`: the same in `verifyWorkflowSecret` and in the generated Nitro workflow runtime (`verifySecret`)

`===` on strings short-circuits at the first differing character, so it leaks (through timing) how many leading characters of a guessed secret are correct. It is the one secret gate in each subsystem.

## Change

A pure-JS constant-time comparison, `timingSafeStringEqual` (new `secret-compare.ts`), used at all three sites. It always inspects every character of equal-length inputs and returns false on a length mismatch (the token length is a fixed, non-secret property). Pure JS on purpose: cron and workflow routes deploy to every runtime, including edge runtimes like Cloudflare Workers where `node:crypto` is unavailable, so `crypto.timingSafeEqual` is not an option there. The generated workflow runtime gets an inline copy of the same routine (it already inlines its other helpers).

## Safety and compatibility

- Behavior is identical for a correct secret (still authorized) and a wrong one (still 401); only the timing side channel is closed.
- No config or API change. The rest of the cron/workflow auth path (case-insensitive header lookup, fail-closed in production when the secret is unset) is untouched.

## Verification

- New unit test `secret-compare.test.ts`: equal strings match, any content difference (including a correct prefix) and any length mismatch return false without throwing.
- Existing `cron.test.ts` and `workflows.test.ts` secret tests (authorized Bearer, unauthorized, whitespace) pass unchanged, exercising the new comparison in both the source and generated paths.
- `pnpm --filter @farm.js/core type-check` passes; the full suite's only failures are the known sibling-dist integration and parallel-build tests, none cron/workflow related.

---
## Summary by cubic
Replaces the `===` secret checks in cron and workflow request auth with a constant-time string comparison so attackers can no longer infer a guessed secret's correct prefix from response timing.

- Adds a pure-JS `timingSafeStringEqual` helper (no `node:crypto`) so the fix works on edge runtimes like Cloudflare Workers.
- Applies it in cron's `isCronRequestAuthorized` and both workflow verifiers; the generated workflow runtime gets an inline copy of the routine.
- Correct secrets still authorize and wrong ones still return 401; only the timing side channel is gone.
- Adds `secret-compare.test.ts` covering identical, differing, prefix, and length-mismatch cases; existing cron and workflow auth tests pass unchanged.

<sup>Written for commit 62e58c8dab7ed9087e6c6fe1e0b47ec6e1583df7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1136?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
superseded by #1130, which landed the same constant-time cron/workflow secret comparison (farmSecretsMatch) while this was in review. closing as redundant.

PR #1140, opened 2026-09-17
fix(plugin): wait for runtime startup before shutting down

Kinfe123:
## Problem

`closeRuntime()` does not wait for an in-progress `startRuntime()`, so stopping a runtime during startup races it. This is reachable in ordinary use: `startDevServer` calls `startRuntime()`, and a dev server stopped while plugins are still setting up triggers `closeRuntime` from the `httpServer` close listener.

Two observable failures, both reproduced:

```
runs shutdown hooks only after startup finishes
  → expected [ 'shutdown', 'setup' ] to deeply equal [ 'setup', 'shutdown' ]

disposes a resource registered while shutdown is already requested
  → Farm runtime cleanup cannot be registered after shutdown begins
```

The first shows a plugin's `shutdown` hook running before its own `setup` completed. The second is the leak: a plugin that opens a database or queue in `setup` and registers the disposer afterwards cannot register it at all, so the resource stays open for the life of the process.

## Root cause

```ts
async closeRuntime(reason = "runtime-closed"): Promise<void> {
  if (this.runtimeClosePromise) return this.runtimeClosePromise;
  if (this.runtimeClosed) return;

  this.runtimeClosePromise = (async () => {
    // went straight to shutdown hooks and then drained the disposer list
```

Nothing consults `this.runtimeStartPromise`, which the class already tracks. Because `runtimeDisposers.splice(0)` drains the list exactly once, any registration after that moment is silently dropped.

The registration guard made the second failure worse rather than catching it:

```ts
if (this.runtimeClosePromise || this.runtimeClosed) {
  throw new Error("Farm runtime cleanup cannot be registered after shutdown begins");
}
```

`runtimeClosePromise` is assigned as soon as `closeRuntime` is called, before any shutdown work happens, so the guard fires during the race window and turns a leak into a thrown error inside plugin setup.

No test covered it because the existing lifecycle tests await `startRuntime()` before closing, which is the sequence that never races.

## Change

Two coupled edits in `plugin.ts`:

1. `closeRuntime` awaits an in-progress `runtimeStartPromise` before running shutdown hooks or draining disposers. A startup failure is swallowed there deliberately: `startRuntime` already reports it to its own caller, and shutdown still needs to run.
2. The `lifecycle.onShutdown` guard now keys off a new `runtimeShutdownStarted` flag, set once disposal is genuinely underway, instead of `runtimeClosePromise`. Setup finishing inside the wait window can register its cleanup; registration after disposal begins is still rejected.

## Safety and compatibility

- The protection the guard was written for is preserved, and is covered by a test that passes both before and after this change, so it is not simply relaxed.
- Shutdown ordering is unchanged otherwise: hooks first, then disposers in reverse registration order, with errors aggregated into `FarmRuntimeShutdownError`.
- `closeRuntime` remains idempotent and still returns the same in-flight promise to concurrent callers.
- No public API or configuration change.
- **Known limit:** a plugin that calls `closeRuntime()` from inside its own `ready`/`setup` hook now awaits the startup it is part of, which would deadlock. That is a pathological pattern, and previously it produced the silent race this PR fixes rather than working correctly. Worth a guard if it ever shows up in practice.

## Verification

macOS, Node v23.11.0, pnpm 8.12.1, against `main` at `a18cce8b`.

```bash
pnpm --filter @farm.js/core exec vitest run src/__tests__/plugin-runtime-shutdown-race.test.ts
pnpm --filter @farm.js/core type-check
pnpm exec oxlint packages/farm/src/plugin.ts
```

3 new tests pass. The neighbouring lifecycle suites pass unchanged: `plugin-lifecycle` (13), `production-lifecycle` (11), `client-lifecycle-hooks` (12), `plugin-routes` (8).

**Failing without the fix.** Reverting `plugin.ts` and keeping the tests fails 2 of 3, with the two messages quoted above. The third test — cleanup rejected once disposal is underway — passes in both states, which is the control proving the guard still works.

`plugin-routes-production.test.ts` fails on this machine both with and without the change (`Build failed with 3 errors`, needing sibling package dists a partial local build does not produce); verified by running it on a stashed clean tree.

## Documentation and release

None. Internal lifecycle ordering fix with no user-facing surface.

---
## Summary by cubic
Fixes a shutdown/startup race so plugins that register cleanup before or during shutdown no longer leak resources or throw.

**Bug Fixes**
- `closeRuntime` no longer waits for an in-progress `startRuntime`, so a plugin whose `runtime.start()` hangs can still be stopped.
- Disposers are drained in a loop instead of spliced once, so cleanup registered while shutdown is running still executes.
- Registering `onShutdown` after the runtime is closed runs it immediately instead of throwing.

<sup>Written for commit 10352b8673f70f85edbaa568d4f9c5b3ab0ae752. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Force-pushed a reworked fix. CI caught that my first attempt broke `shuts down when SIGTERM interrupts runtime startup` (1 failed of 2364), and it was right to.

That test's plugin has a `runtime.start()` that never resolves:

```js
async start() { await new Promise(() => setInterval(() => {}, 1_000)); }
```

Awaiting startup inside `closeRuntime` therefore deadlocks, the graceful timeout force-exits, and the process returns 1 instead of 0. Shutdown not waiting for startup is a deliberate contract, so the original approach was wrong regardless of the leak it fixed.

The rework keeps shutdown prompt and makes disposal resilient instead:

- the disposer list is drained in a loop instead of `splice(0)` once, so cleanup registered while draining still runs
- a resource registered after the runtime has fully closed is disposed immediately instead of throwing back into plugin setup, which is the only outcome that neither leaks nor breaks setup

New test `does not wait for a startup that never finishes` encodes the contract I violated, so this cannot regress the same way. Note that `production-prebuilt-ssr.test.ts` cannot be validated locally on this machine — it fails 32/33 on upstream main here for want of built sibling packages — so CI is the real check for that file.

PR #1122, opened 2026-09-17
perf(react): reuse owned keyed reorder bookkeeping

Kinfe123:
## Problem

After a validated compiled keyed reorder, the block copies an already-owned row map and rebuilds every element-to-row lookup, even when all row objects and root elements survive. A cancelled double reversal also does this unnecessary O(n) work.

This follows main independently of #1118. The shared array-based map-building helper, benchmark source, controls and thresholds remain unchanged. **Merge decision:** the maintainer has requested proceeding with squash merge after the incomplete performance verification was explained. Correctness/compatibility/size checks and CI pass, but equal-or-better browser performance is **not verified**. This is not a benchmark pass or a speedup claim.

## Root cause

The private reorder contract exposed only ReadonlyMap despite returning a native map owned by this block. The consumer copied it and rebuilt the element index without using the existing retained-row invariant. All native, mapped and structural reorder implementations return permutations/subsets of existing row instances; they do not create or replace root elements. Updated item/index fields already live on those same instances.

## Change

- Express native-map ownership and retained-row identity in the private reorder contract, without casts or public API changes.
- Adopt the owned row map directly.
- Retain the element lookup when every row survives. When rows are removed, still rebuild it and prune handlers/listeners in the existing removal branch.
- Add ten regression cases across static/hybrid modes: repeated single/double/triple reversals, mapped edits followed by reversal, and structural removals. Check map-copy/index-write counts, exact DOM moves, row identity/content, delegated clicks receiving current item values/indexes, and no owner reruns.
- Update the experimental compiler docs and benchmark example README.

## Safety and compatibility

Eligibility checks, native method evaluation/errors, preparation before DOM mutation, LIS/reverse moves, focus/selection, removal cleanup, collection commits, afterCommit and complete fallbacks are unchanged. Reorder results are private runtime-owned maps; apps cannot supply a custom ReadonlyMap through this interface. Append, position, rolling-window and general reconciliation logic are unchanged.

No new configuration, syntax, dependencies, versions, release changes or renderer-neutral/core changes. Compiler-off and direct-component bundle fixtures are unchanged. The retained-instance requirement is documented at the private interface to constrain future reorder implementations.

## Verification

[...]

PR #1142, opened 2026-09-17
fix(react): reject invalid benchmark options before trials

Kinfe123:
## Problem

The compiler dashboard benchmark accepts invalid numeric environment options and starts expensive trial work before discovering the error. For example, `FARM_TABLE_SAMPLES=NaN`, negative warmup and a base port of 65533 currently reach browser startup. Non-finite or fractional sample counts can skip or distort loop execution; the fourth trial's port can exceed 65535.

## Root cause

The runner converts six environment options with `Number(...)` without validating their bounds or integer semantics. Report preflight protects output files, but does not validate the experiment configuration.

## Change

- Read and validate the existing numeric options before report creation, browser launch or production-trial build work.
- Require positive safe-integer sample/update/cycle counts, nonnegative safe-integer warmup, and a base port in 1–65532 for four consecutive trial ports.
- Name the offending option in the error. Do not clamp, round or silently replace invalid values.
- Keep existing defaults, empty/unset behavior, valid numeric spellings and explicit zero warmup. Document the supported bounds in the example README.

## Safety and compatibility

Only the benchmark runner changes. No framework/compiler/runtime code, public config flag, dependency, bundle output or release version changes.

All benchmark code after option initialization is byte-for-byte unchanged: workload, sample defaults, trial order, timers, native controls, aggregation, performance thresholds and artifact publication. One-sample diagnostic overrides remain accepted; this validation does not make reduced or noisy runs sufficient for performance claims. It does not check port availability or promise that the host has enough resources.

The package's existing `prebenchmark` renderer build still happens before this script; validation precedes the script's production trial work.

## Verification

Local Darwin arm64 with Node23.11.0 and Node24.19.0:

- `pnpm --filter @farm.js/react exec vitest run src/__tests__/compiler-dashboard-options.test.ts src/__tests__/compiler-dashboard-artifacts.test.ts src/__tests__/compiler-dashboard-controls.test.ts --fileParallelism=false`: 94 pass (70 options tests, 10 artifact tests, 14 unchanged compiler/native-control tests).
- The same 94 tests pass on Node24.19.0 by invoking `node node_modules/vitest/vitest.mjs run ... --fileParallelism=false` from `packages/farm-react`.
- Three child-process tests launch the real runner with invalid configuration and an unavailable browser. They require option-specific failure before any artifact directory, browser or trial build and verify that the previous report and build sentinel remain untouched.
- The report-path regression test now explicitly uses numeric defaults so unrelated shell options cannot mask its filesystem assertion. All 94 focused tests also pass with `FARM_TABLE_SAMPLES=invalid FARM_BENCHMARK_WARMUP=-1` set in the invoking shell.
- Regression proof: temporarily restored the six old unchecked initializers; all three child-process tests fail because execution creates trial artifacts and reaches browser launch. The mutation was removed and the complete focused suite passes.
- `pnpm --filter @farm.js/react type-check`, Node syntax checks, focused oxfmt/oxlint, and `git diff --check` pass.
- Measurement-function SHA256 stays `87d7a665f60eefb8f7b6bdbd745af30b6324a35a721db19da5923b5073513d58`; aggregation/gates/report-object SHA256 stays `7fe99adce48e755f06d0bc202cce394a61bea7715417b11436075cb8afaa6f5c`.

No new browser performance result is claimed. The earlier #1122 before/after comparison remains unverified; this follow-up does not replace it. No browser timing run was performed for this tooling-only follow-up.

## Documentation and release

Updated `examples/react-compiler-dashboard/README.md`. No docs routes, generated artifacts, templates, package versions, tags or releases changed.

---
## Summary by cubic
Rejects invalid numeric benchmark options in the compiler dashboard runner before it creates reports, launches a browser, or starts production-trial builds. Previously values like `FARM_TABLE_SAMPLES=NaN`, negative warmup, or `FARM_DASHBOARD_PORT=65533` reached browser startup and could skew or skip loop execution.

- Requires positive safe-integer sample/update/cycle counts, nonnegative safe-integer warmup, and a base port from 1–65532 so all four trial ports stay valid.
- Throws a `TypeError` naming the offending option; invalid values are rejected, not clamped or rounded.
- Keeps existing defaults, empty/unset handling, valid numeric spellings, and explicit zero warmup; only the benchmark runner changes, and the README documents the new bounds.

<sup>Written for commit b2b707d0936516f5d3c139dd4306818190adeaa6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1142?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Merge verification for b2b707d0936516f5d3c139dd4306818190adeaa6: all GitHub CI jobs passed, including Windows, React 18/19 compatibility, builds and E2E suites. The Farm preview deployment succeeded. The docs preview reports "Deployment rate limited — retry in 24 hours"; the maintainer explicitly authorized merging with this rate-limit exception. This is not a successful docs deployment or a reported build failure. There are no review threads; Cubic returned neutral/skipped rather than an approval. Proceeding with a squash merge.

PR #1145, opened 2026-09-17
fix(react): stop benchmark servers after browser setup failures

Kinfe123:
## Problem

Follow-up to #1142.

A failed compiler dashboard trial can leave its production server running. The server is spawned before browser setup, but browser context creation, page creation and event registration happen before the cleanup `try`. Separately, a rejected `context.close()` prevents the subsequent `stopServer()` call.

The regression fixture injects those browser failures into the actual runner function. With the old code, real child processes remain alive after the trial rejects, potentially occupying resources and interfering with later benchmark attempts.

## Root cause

The teardown boundary does not cover acquisition of the browser resources, and the two cleanup operations are sequential without protecting server shutdown from a context-disposal failure.

## Change

- Move context/page setup into the existing trial `try`, keeping the context reference available for cleanup.
- Close the context only if it was acquired, and put server shutdown in its own `finally` so it is attempted even when context setup/disposal fails.
- Add focused failure-injection coverage of the actual runner function, including three cases with real child processes and the unchanged production shutdown helper.
- Document the failure cleanup behavior in the maintained example README.

## Safety and compatibility

Only the benchmark runner's cleanup boundary changes; no compiler, React runtime, framework API, bundle output or dependency changes.

The successful measurement path, browser settings, warmup, workloads, sample counts, controls, summaries and gates are unchanged. The code from server readiness through the returned trial result is byte-for-byte identical. Shutdown behavior/timeouts and top-level trial sequencing/report publication are unchanged. Failures still reject the trial and abort the run; cleanup follows ordinary JavaScript `finally` error propagation, without retries or converting failures into successful reports.

This is not a general process supervisor or a new cancellation/timeout system. Tests use a deterministic stdout readiness event for the real child fixtures, not a fixed sleep. Test cleanup tracks only the children it owns, including when running negative controls.

## Verification

Local Darwin arm64, Node23.11.0 and Node24.19.0:

- Focused dashboard suites: `pnpm --filter @farm.js/react exec vitest run src/__tests__/compiler-dashboard-lifecycle.test.ts src/__tests__/compiler-dashboard-options.test.ts src/__tests__/compiler-dashboard-artifacts.test.ts src/__tests__/compiler-dashboard-controls.test.ts --fileParallelism=false`.
- 110 tests: 16 lifecycle cases plus the existing 94 option/report/native-control tests. The same selection passes with Node24.19.0 using `node packages/farm-react/node_modules/vitest/vitest.mjs run --config packages/farm-react/vitest.config.ts` followed by those four repository-relative test paths and `--fileParallelism=false`.
- Lifecycle cases cover failure before resource acquisition, context/page/event setup, server readiness, navigation, measurement, screenshot, context disposal, shutdown rejection and the successful settings/teardown order.
- Three tests start a real Node child and assert it has exited after context creation, page creation or context disposal fails. Browser behavior is fault-injected; these are process-lifecycle integration tests, not a full browser performance run.
- Regression proof: temporarily restored the old cleanup placement. Eight regression cases failed, including all three real-process cases because the child was still alive. The test teardown reaped those owned children; the mutation is not included.
- `pnpm --filter @farm.js/react type-check`, focused oxfmt/oxlint, Node syntax check and `git diff --check` pass.
- Unchanged readiness/measurement/result block SHA256: `7fb34b2339cf1e5af84bf8632557c1957b421f8950d6111785aa3fae3659510e`.
- Unchanged aggregation/gates/report-object SHA256: `7fe99adce48e755f06d0bc202cce394a61bea7715417b11436075cb8afaa6f5c`.

No new performance result is claimed. #1122's interrupted before/after performance comparison still needs an uninterrupted run on a suitable host; cleanup tests do not replace it.

## Documentation and release

Updated `examples/react-compiler-dashboard/README.md`. No docs routes, generated output, templates, package versions, tags or releases changed.

---
## Summary by cubic
Stops the benchmark trial's production server even when browser setup or context disposal fails, so failed trials no longer leave orphaned processes running.

- Moves context and page setup inside the cleanup `try` so acquired resources are covered by teardown.
- Guards `context.close()` on whether the context was acquired and runs server shutdown in its own `finally`.
- Adds lifecycle tests that inject failures into the actual runner function, including real child-process cases that verify the server is reaped.
- Documents the cleanup behavior in the example README.

<sup>Written for commit 73164cdede07fb66887a5ab2083eddf27c0e9f9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Merge verification for 73164cdede07fb66887a5ab2083eddf27c0e9f9b: all GitHub CI jobs passed, including Linux/Windows tests and type-checks, React 18/19, builds and E2E suites. Both Vercel previews report "Deployment rate limited — retry in 24 hours"; the maintainer explicitly authorized the deployment rate-limit exception. This does not represent successful deployments or verified deployment builds. There are no review threads; Cubic is neutral/skipped, not an approval. Proceeding with a squash merge.

PR #1147, opened 2026-09-17
fix(react): confirm benchmark server exit before continuing

Kinfe123:
## Problem

Follow-up to #1145.

The compiler dashboard benchmark can return from server cleanup before its child process has exited. A server that ignores `SIGTERM` receives `SIGKILL` after the existing two-second grace period, but the old helper immediately returns instead of awaiting the exit event. Subsequent trial work can therefore overlap the old server's lifetime.

The regression fixture starts a real Node child that ignores `SIGTERM` on POSIX. Before the fix, `stopServer()` resolves while both exit indicators are still null; after the fix, it resolves only after the child exits. Windows exercises its native immediate `SIGTERM` termination behavior; deterministic event tests cover escalation on every platform.

## Root cause

The helper treats signal delivery as shutdown completion. It also checks only `exitCode`, although signal termination leaves that null, installs its exit listener after the first signal, and leaves the grace timer running after early exit. [Node's child-process contract](https://nodejs.org/api/child_process.html#subprocesskillsignal) distinguishes sending a signal from the child actually exiting.

## Change

- Recognize either an exit code or a terminating signal as an already-exited child.
- Install owned exit/error listeners before sending `SIGTERM`.
- Preserve the existing two-second escalation to `SIGKILL`, but continue waiting for a confirmed exit.
- Reject signal-delivery errors, thrown failures and false return values instead of reporting successful cleanup.
- Remove only this shutdown operation's timer and listeners on completion or error.
- Extend the existing lifecycle tests and example README.

## Safety and compatibility

Benchmark tooling only: no compiler/runtime, framework API, dependency, bundle, release or generated-output changes. The complete trial implementation, workload timings, samples, controls, report aggregation and gates are unchanged. Teardown happens outside measured browser updates.

No new grace interval, retry policy, process-tree supervisor or benchmark threshold. Only the child owned by the trial is signaled. Signal-delivery failure aborts the run; it does not continue under an assumption that shutdown succeeded. After successful forced-signal delivery, cleanup waits for the OS exit event rather than adding a deadline that can falsely report success. This does not add general benchmark cancellation or promise to stop descendant processes.

## Verification

Local Darwin arm64, Node23.11.0 and Node24.19.0:

- `pnpm --filter @farm.js/react exec vitest run src/__tests__/compiler-dashboard-lifecycle.test.ts src/__tests__/compiler-dashboard-options.test.ts src/__tests__/compiler-dashboard-artifacts.test.ts src/__tests__/compiler-dashboard-controls.test.ts --fileParallelism=false`: 123 passing tests, including 29 lifecycle tests and the existing 94 option/report/native-control tests.
- Same 123 tests passed on Node24 using its `node` executable with `packages/farm-react/node_modules/vitest/vitest.mjs run --config packages/farm-react/vitest.config.ts`, the four repository-relative paths above and `--fileParallelism=false`.
- Thirteen new tests cover already-exited and signaled children, graceful exits, delayed forced exits, listener-before-signal ordering, emitted/thrown/false-return signal failures, owned-listener/timer cleanup, and a real stubborn child. Existing real-child browser failure tests still pass.
- Negative control: new tests against the old helper produced 11 regression failures, including the real-process assertion that shutdown had returned before exit; 18 existing/control cases passed. No unhandled rejections in the final negative-control run. Owned test children were reaped by teardown.
- `pnpm --filter @farm.js/react type-check`, focused `oxfmt`/`oxlint`, Node syntax check and `git diff --check` pass.
- Complete trial implementation before the artifact setup is byte-for-byte unchanged: SHA256 `c386a23c02e01296fcdbd6f15801b86c37f1ec7fb7d1162229aaeeaf5a573772`.
- Readiness/measurement/results SHA256 remains `7fb34b2339cf1e5af84bf8632557c1957b421f8950d6111785aa3fae3659510e`; aggregation/gates/report-object SHA256 remains `7fe99adce48e755f06d0bc202cce394a61bea7715417b11436075cb8afaa6f5c`.

These are lifecycle and correctness checks, not a new browser timing result. No speedup or measured nonregression is claimed. #1122's interrupted before/after performance comparison still needs an uninterrupted run on a suitable host.

## Documentation and release

Updated `examples/react-compiler-dashboard/README.md`. No docs routes, generated output, templates, versions, tags or releases changed.

---
## Summary by cubic
Fixes the compiler dashboard benchmark's server shutdown so `stopServer()` waits for the child process to exit before the next trial starts. It previously returned once `SIGTERM` was sent, which let a slow-to-exit server overlap the next trial; now it waits through the two-second `SIGKILL` escalation and resolves only after the exit event. Signal-delivery failures now abort the run instead of reporting successful cleanup.

**Bug Fixes**
- Skips signaling when the server has already exited, via exit code or signal code.
- Installs exit and error listeners before signaling and removes only its own timer and listeners on completion.
- Rejects thrown or false `kill()` returns as shutdown failures.
- Adds lifecycle tests for already-exited, graceful, escalated, and stubborn-child shutdowns, and updates the example README.

<sup>Written for commit 9a09a163beac9667c37802d518aa97a0e0ec4755. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Merge verification for 9a09a163beac9667c37802d518aa97a0e0ec4755: all 14 GitHub CI jobs passed, including Linux/Windows tests and type-checks, React 18/19, builds and E2E suites. Both Vercel previews report "Deployment rate limited — retry in 24 hours"; merging under the maintainer-authorized rate-limit exception, not treating those deployments as successful. No review threads; Cubic returned neutral/skipped rather than an approval. Proceeding with a squash merge.

PR #1200, opened 2026-09-18
fix(framework): resolve reviewed runtime regressions

Kinfe123:
## Problem

The remaining automated follow-up PRs each addressed an isolated regression, but merging them independently would have left overlapping changes and several correctness gaps in production paths. The affected behavior spans request handling, routing, integration middleware, renderer streams, build orchestration, generated metadata, and recovery/lifecycle behavior.

This consolidates and supersedes #1159 and #1167 through #1197.

## Root cause

Most failures were edge cases outside the original happy-path coverage: decoded zstd bodies retaining framing headers, cache keys collapsing distinct web-platform values, recursive matchers backtracking exponentially, refreshed cookies being overwritten between middleware, catalog watchers matching locale prefixes, and Nitro starting from per-environment hooks before all Vite outputs were current.

The individual RSC fix also treated `buildApp` as an ordinary plugin hook on Vite 6. In practice, `@vitejs/plugin-rsc` owns the application builder there, so the callback was replaced and a plain `vite build` did not create Nitro output.

## Change

- Preserve binary and multipart API bodies, route failures, middleware ordering, lifecycle failure responses, preload detection, and chunk recovery semantics.
- Harden route scanning and production matching, including bounded catch-all backtracking and regex statements after control-flow conditions.
- Canonicalize cache keys without losing repeated-value order and handle RFC 8878 skippable zstd frames.
- Preserve accumulated Supabase refresh cookies and correct auth, billing, Stripe storage, WorkOS, workflow, and webhook edge cases.
- Fix Vue stream termination, renderer style compatibility, iframe accessibility hints, i18n catalog resolution, metadata images, and docs TOC generation.
- Compose Nitro after the complete Vite RSC application build on Vite 6 and through the application hook on Vite 7+, while preserving custom `assetsDir`.
- Add focused regression coverage for each corrected mechanism.

## Safety and compatibility

- No intentional breaking public API changes.
- React remains the default renderer; React, Preact, Solid, Vue, and Svelte renderer contracts are covered.
- Server-only provider state and secrets remain outside client entry points.
- Vite development and built Nitro paths retain native Request/Response, cookie, header, and body semantics.
- Nitro remains opt-in through the existing `nitro()` plugin and runs once per completed application build.

## Verification

- Core suite: 2,558 tests passed; the 11 suites initially blocked by unbuilt workspace package entries passed 40/40 after building those entries.
- `pnpm --filter @farm.js/plugin test` — 159/159
- `pnpm --filter @farm.js/auth test` — 10/10
- `pnpm --filter @farm.js/cli test` — 123/123
- `pnpm --filter @farm.js/hints test` — 19/19
- `pnpm test:renderers` — all React, Preact, Solid, Vue, and Svelte tests and type checks passed
- Docs waitlist route regression: 9/9
- Package type checks/builds passed for core, plugin, auth, CLI, hints, Supabase, Autumn, and Stripe.
- `NEXT_PUBLIC_SUPABASE_URL=https://example.supabase.co NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY=test-key pnpm --dir examples/supabase-integration build`
- `NITRO_PRESET=node-server pnpm --dir examples/rsc-demo exec vite build`; booted `.output/server/index.mjs` and verified `GET /` returns HTTP 200.
- Focused `oxfmt`, `oxlint`, and `git diff --check` completed with no errors.

## Documentation and release

Includes the reviewed documentation/runtime corrections and new regression fixtures. No version or release metadata is changed.

---
## Summary by cubic
Consolidates reviewed runtime regression fixes across request handling, routing, middleware, integrations, renderers, and the Nitro build pipeline. No intentional breaking public API changes.

**Request handling, routing, and data integrity**
- Preserves decoded zstd, binary, and multipart API bodies without stale content metadata.
- Replaces regex-based route scanning with a tokenizer-aware scanner and bounds catch-all matching to prevent phantom routes and exponential backtracking.
- Canonicalizes cache keys while keeping repeated-value order; the API client no longer wipes its cache on transient header failures.
- Forwards integration middleware `Set-Cookie` values to the final response and preserves Supabase refresh cookies.
- Auth0 and WorkOS redirect unauthenticated requests with 303 so the GET-only login route is reached regardless of method.
- Autumn marks expired purchases as canceled; Stripe Prisma storage drops the exact unknown field Prisma reports; workflow routes verify the secret before the id lookup.
- Failed runtimes return a controlled 503 with no retry-after; chunk recovery retries when storage is unavailable; error diagnostics redact additional secret patterns.

**Renderers, i18n, docs, and build**
- Vue, Solid, and Svelte renderers match React's unitless style set and ignore boolean, null, and undefined children.
- Vue mid-stream errors end the destination cleanly instead of crashing the worker.
- Metadata images resolve `React.lazy` components; preload scanning handles self-closing SVG elements.
- The i18n catalog watcher no longer matches locale prefixes, and base languages are preferred over regional variants.
- Docs Setext headings appear in the TOC; docs fonts and configured headers apply to the locale-stripped pathname.
- Nitro builds after the complete Vite RSC application build on Vite 6 and via the `buildApp` hook on Vite 7+.
- RSC client navigation ignores modified and non-primary clicks.

<sup>Written for commit 703004cccbd30982ad3042506d543d5f39298b30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
Closing this consolidated version so each reviewed fix can be submitted as its own focused pull request.

PR #1237, opened 2026-09-18
feat(cli): support host binding in farm dev

Kinfe123:
## Problem

`farm dev` printed Vite's `use --host to expose` guidance but rejected `--host` as an unknown CLI option, so apps could not opt into LAN/container binding from the command line.

## Root cause

The CLI exposed only `--port`, and the core `startDevServer` helper had no host override to apply after loading the project's Vite configuration.

## Change

- Add Vite-compatible `-H, --host [host]` handling to `farm dev`.
- Thread string and boolean host values through the CLI and core dev-server helpers.
- Make an explicit CLI host override the host in `farm.config.ts` before listening.
- Print Vite's resolved local and network URLs in the Farm-branded startup output.
- Document both automatic network exposure and explicit host binding.

## Safety and compatibility

Existing calls to `startDevServer(config, port)` are unchanged; the host is a new optional third argument. Without `--host`, binding behavior remains controlled by project configuration and Vite defaults. `--host` without a value uses Vite's boolean exposure behavior, while `--host 127.0.0.1` binds the named host.

## Verification

- `pnpm --filter @farm.js/cli test` (124 passing)
- `pnpm --filter @farm.js/cli type-check`
- `pnpm --filter @farm.js/core test -- src/__tests__/development-host.test.ts`
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm --filter @farm.js/core build`
- The regression fixture configured `0.0.0.0`, started with an explicit `127.0.0.1` override, verified the bound socket address, and completed a real HTTP request.
- Focused `oxfmt`, `oxlint`, and `git diff --check`

## Documentation and release

Updated the CLI command reference. The next shared core/CLI beta publishes the option.

---
## Summary by cubic
Adds Vite-compatible `--host` support to `farm dev` so apps can bind the dev server to a specific host or expose it on the local network from the CLI. Previously `--host` was rejected as an unknown option, and startup output always showed the static "use --host to expose" hint.

- `-H, --host [host]` overrides the project's configured host; using `--host` with no value enables Vite's network-exposure behavior.
- The host override is applied after project config loads; existing `startDevServer(config, port)` callers keep their current behavior.
- Startup branding now prints Vite's resolved local and network URLs, showing real network addresses when exposure is on.
- Updated the CLI command reference to document both automatic network exposure and explicit host binding.

<sup>Written for commit 2c11920924b9e46bbbd34fa54cf2a98fe53dc9d3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/farming-labs/farm.js/pull/1237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kinfe123:
closing as superseded: farm dev already has -H/--host on main (packages/farm-cli/bin/farm.js line 137), and #1366 just fixed the override to apply during config resolution instead of after server creation. this branch's conflicts are against that landed version, so there is nothing left to take from it.

